PR #5, opened 2021-03-31
statement: fix compilation on 64-bit Linux

pborzenkov:
The patch adjusts the types used to communicate with DuckDB C API to use
exactly the same datatypes as duckdb.h uses.

This fixes compliation for me which failed before the change with the
following error:

```
/go/pkg/mod/github.com/marcboeker/go-duckdb@v0.0.0-20210302143741-5b815576076f/statement.go:41:32:
  cannot use &pc (type *_Ctype_ulonglong) as type *_Ctype_ulong in assignment

<similar error>
```

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -37,7 +37,7 @@ func (s *stmt) NumInput() int {
 	if s.closed {
 		panic("database/sql/driver: misuse of duckdb driver: NumInput after Close")
 	}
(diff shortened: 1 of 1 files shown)


marcboeker:
@pborzenkov Awesome, thank you for this fix!

Issue #7, opened 2021-11-15
 field undefined 


liubaochuan:
../../go/pkg/mod/github.com/marcboeker/go-duckdb@v0.0.0-20210403084307-05501463909f/rows.go:70:12: val.year undefined (type _Ctype_struct___3 has no field or method year)
../../go/pkg/mod/github.com/marcboeker/go-duckdb@v0.0.0-20210403084307-05501463909f/rows.go:71:19: val.month undefined (type _Ctype_struct___3 has no field or method month)
../../go/pkg/mod/github.com/marcboeker/go-duckdb@v0.0.0-20210403084307-05501463909f/rows.go:72:12: val.day undefined (type _Ctype_struct___3 has no field or method day)
../../go/pkg/mod/github.com/marcboeker/go-duckdb@v0.0.0-20210403084307-05501463909f/statement.go:41:18: too many arguments in call to _Cfunc_duckdb_nparams

duckdb version: 0.3.0

marcboeker:
Merged #8 to fix this. Thanks @mesge!

PR #8, opened 2021-11-18
Use duckdb_date_struct instead of duckdb_date

mesge:
As described in #7 `rows#Next` tries to access undefined fields on a struct. Looks like the breaking change was introduced upstream in https://github.com/duckdb/duckdb/pull/2107.

This PR updates `rows#Next` from `duckdb` package to use the [`duckdb_date_struct`](https://github.com/duckdb/duckdb/blob/8ef7ee22c543fb030c0e6e62b5918dd43cc46fc7/src/include/duckdb.h#L81), which contains the fields that this driver expects.


marcboeker:
@mesge Thanks for submitting this PR to fix the breaking change :)

PR #13, opened 2022-03-14
Add support to scan hugeint type

ajzo90:
Needed for SUM(intColumn) to work


marcboeker:
Thanks for adding HugeInt support!

PR #14, opened 2022-03-20
fix(date): updated code to convert to correct struct

wolfeidau:
Looks like at some point the structure returned in rows went back to `duckdb_date` therefore dates were all messed up.

I added a test to pick up this regression in the future.


marcboeker:
Thanks for the fix 👍

PR #15, opened 2022-03-22
chore(project): add github action for tests

wolfeidau:
Just adding a simple `Makefile` to help with feedback on PRs

Open to suggestions on how this works, I tried to add a macosx runner but the version of OSX isn't compatible with the binaries published by duckdb project.

Hopefully this changes when github releases newer agents, see https://github.com/github/roadmap/issues/240

### DIFF
--- a/.github/workflows/cgo.yaml
+++ b/.github/workflows/cgo.yaml
@@ -1,0 +1,29 @@
+name: Go
+
+on: [push, pull_request]
+
+jobs:
+
+  test:
+    name: Test
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        go: ['1.17']
+      fail-fast: false
+    
+    steps:
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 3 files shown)


marcboeker:
Wow this is awesome. Thanks!

When adding `-ldflags="-r $(LIB_PATH)"` to the `go run` or `go test`  I was able to make it work. Could you please check if the following solves the problem under macOS for you?

```
.PHONY: build
build: $(LIBS)
	$(LDFLAGS) go run -ldflags="-r $(LIB_PATH)" examples/simple.go

.PHONY: test
test: $(LIBS)
	$(LDFLAGS) go test -ldflags="-r $(LIB_PATH)" -v ./...
```

wolfeidau:
@marcboeker I need to read up more on these flags, I have tested the existing code on OSX 12 (Monteray) and Ubuntu 22.10 and it was 👍🏻 

That said the changes you suggested don't break my local builds so I have pushed them, having them work for everyone is the goal.

I have also verified this is fine in GitHub Actions as you can hopefully see here https://github.com/wolfeidau/go-duckdb/runs/5655482453?check_suite_focus=true.

marcboeker:
@wolfeidau I've tested `macos-latest` and it works on the runner. Please see https://github.com/marcboeker/go-duckdb/runs/5661424780?check_suite_focus=true

So I'm going to merge these changes and add macOS to the list of platforms to be tested.

PR #21, opened 2022-04-24
fix some bugs

junhwong:


statement_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"testing"
+)
+
+func TestPrepareQuery(t *testing.T) {
+
+	db := openDB(t) //:memory:
+
+	defer db.Close()
+
+	stmt, err := db.Prepare("select * from tbl where col=?")
```
marcboeker, line 13: @junhwong The tests are failing, as the specified table is not available. Maybe you can use the [createTable()](https://github.com/marcboeker/go-duckdb/blob/master/duckdb_test.go#L194) function beforehand and the select from the `foo` table.

marcboeker (commented):
Thanks for the PR. Please see comment for the failing tests.

PR #25, opened 2022-06-13
upgrade duckdb, add variable, macos universal, add race detector

segfault88:
- Upgrade duckdb version in Makefile to 0.3.4 and add variable for future updates
- For Mac, use the universal build (support Intel and Apple / Aarch64 / ARM64 CPUs)
- Add race detector to test run


segfault88:
Thanks, I have rebased on your upstream now - hopefully that works properly now.

PR #33, opened 2022-07-26
Decimal Type Support

### LINKED ISSUE #32: Support for decimal types
Scanning of decimal is not currently implemented.
If we try to scan a decimal we get the error `not supported type 19`.

Decimal types come up pretty often. Something as simple as `select 1.0 * 1.0` yields a decimal result.

andyyu2004:
Resolves #32


marcboeker:
Thanks for adding decimal type support!

PR #34, opened 2022-07-28
Add time.Time support for duckdb timestamp type.

panta:


marcboeker:
Thanks for the contribution!

PR #35, opened 2022-07-29
Fixing string with size=12

AdityaHegde:
I was getting segmentation fault when string size was exactly 12.
Needed a <= check for the size in scanBlob


marcboeker:
Thanks for the fix!

PR #37, opened 2022-08-02
Add support for DuckDB Appender API

panta:
Besides the support for the aforementioned [Appender API](https://duckdb.org/docs/api/c/appender), this also

* adds explicit support for various numerical types (UTINYINT, USMALLINT, ...) in particular to add column type support
* add more fuzzing tests for the various basic types
* switch to use duckdb_open_ext() in place of duckdb_open() to be able to return error information


marcboeker:
@panta Thank you for this PR. Please give me some days to review this, as I'm currently quite busy.

panta:
> @panta Thank you for this PR. Please give me some days to review this, as I'm currently quite busy.

Sure, thanks. Just a thing I forgot to mention: tests for the appender use filesystem backed DBs (using temporary files), because the appender API doesn't seem to work with memory DBs.

appender.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"database/sql/driver"
+	"errors"
+	"fmt"
+	"time"
+	"unsafe"
+)
+
+type appender struct {
```
marcboeker, line 16: Wouldn't it make sense to make the `appender` exported? Then it would also appear in the documentation.

appender.go
```diff
@@ -0,0 +1,157 @@
...
+	if a.closed {
+		panic("database/sql/driver: misuse of duckdb driver: double Close of Appender")
+	}
+
+	a.closed = true
+
+	// if state := C.duckdb_appender_flush(*a.appender); state == C.DuckDBError {
+	// 	dbErr := C.GoString(C.duckdb_appender_error(*a.appender))
+	// 	return errors.New(dbErr)
+	// }
```
marcboeker, line 39: Can we remove this part as `duckdb_appender_destroy` should be sufficient?

connection.go
```diff
@@ -208,3 +208,34 @@ func escapeValue(v string) []byte {
...
+
+	var a C.duckdb_appender
+	if state := C.duckdb_appender_create(*c.con, schemastr, tablestr, &a); state == C.DuckDBError {
+		return nil, fmt.Errorf("can't create appender")
+	}
+
+	return &appender{c: c, schema: schema, table: table, appender: &a}, nil
+}
+
+func Appender(dbconn driver.Conn, schema string, table string) (*appender, error) {
```
marcboeker, line 234: I would suggest to add a `NewAppenderFromConn(conn driver.Conn, schema, table string) (*Appender, error)` method to the `appender.go` and move the code from `(c *conn) Appender()` there. This would keep the current DuckDB implementation close to the original `database/sql` design idea and all extensions, like the appender, would be more separated. What do you think about this?

appender.go
```diff
@@ -0,0 +1,157 @@
...
+		case float64:
+			if rv := C.duckdb_append_double(*a.appender, C.double(v)); rv == C.DuckDBError {
+				return fmt.Errorf("couldn't append parameter %d (type %T)", i, v)
+			}
+		case bool:
+			if rv := C.duckdb_append_bool(*a.appender, C.bool(v)); rv == C.DuckDBError {
+				return fmt.Errorf("couldn't append parameter %d (type %T)", i, v)
+			}
+		case []byte:
+			if v == nil {
```
marcboeker, line 119: Isn't this already checked for in line 58?

appender_test.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	dbconn, err := db.Conn(context.Background())
+	require.NoError(t, err)
+	err = dbconn.Raw(func(driverConn any) error {
+		llConn, ok := driverConn.(*conn)
+		require.True(t, ok)
+
+		appender, err := Appender(llConn, "", "appdata")
+		require.NoError(t, err)
+		// defer appender.Close()
```
marcboeker, line 176: Wouldn't this be more safe than a manual `appender.Close()` in line 183?

appender_test.go
```diff
@@ -0,0 +1,197 @@
...
+	return db
+}
+
+func createAppenderTable(db *sql.DB, t *testing.T) *sql.Result {
+	res, err := db.Exec(testAppenderTableDdl)
+	require.NoError(t, err)
+	return &res
+}
+
+func TestAppenderSimple(t *testing.T) {
```
marcboeker, line 35: Maybe we could only use TestRandomizedAppender instead of TestAppenderSimple as it already covers everything that the simple test case does.

appender.go
```diff
@@ -0,0 +1,157 @@
...
+		if v == nil {
+			if rv := C.duckdb_append_null(*a.appender); rv == C.DuckDBError {
+				return fmt.Errorf("couldn't append parameter %d", i)
+			}
+			continue
+		}
+
+		switch v := v.(type) {
+		case uint8:
+			if rv := C.duckdb_append_uint8(*a.appender, C.uint8_t(v)); rv == C.DuckDBError {
```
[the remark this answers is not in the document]
panta: Hey, thanks for the feedback, and I apologize for the delay. In the next few days I'll resume the work on this and address the issues with the PR. Thanks!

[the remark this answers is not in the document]
panta: I've pushed an additional commit that addresses all the points.
If you deem it ok, I could squash the appender commits into one before merging, so we have a cleaner history.
Please let me know.

flarco:
Perhaps I can help get this to a merging state. @marcboeker you stated to split the tests and api changes into two different PRs. I can do that. Anything else you think needs to be done?

marcboeker:
@flarco Sounds great. I think it would make sense to merge the current master into this branch, as there are could be some breaking changes. 

The idea behind splitting the tests was, that some tests are not related to the appender feature. For now I would suggest to exclude non related tests and only keep the appender relates ones in the PR.

flarco:
@marcboeker opened https://github.com/marcboeker/go-duckdb/pull/71 and removed the non-appender related tests. `master` was merged into it and there are no conflicts at the moment. Please review and approve.

Issue #38, opened 2022-08-11
Use DuckDB amalgamation instead of shared libraries


begelundmuller:
Hello! First of all, thank you for building this lovely library!

I wanted to ask if you have considered using CGo to build DuckDB from source instead of downloading and linking the DuckDB shared libraries? We cross-compile and distribute our binaries, so needing to bundle a shared library is quite the headache.

DuckDB is [available](https://duckdb.org/why_duckdb#duckdbissimple) as a single combined C++ source and header file (a so-called "amalgamation"), which should make it easy to bundle together with this repo. That would be similar to what the [go-sqlite3](https://github.com/mattn/go-sqlite3) library does (here's a direct [link](https://github.com/mattn/go-sqlite3/blob/master/sqlite3-binding.c) to the amalgamated SQLite code in the repo).

One caveat is that building DuckDB takes a while (as an example, 10 minutes on my 4-year old MacBook Pro). That would only apply when you first install the library, though. After that Go caches the build (for CI/CD, caching the builds also [appears](https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds) to be reasonably easy).

Overall, assuming this works, the pros/cons would be:

- Pros: gets rid of downloads and build flags (incl. Makefile) and you get a standalone output binary
- Cons: initial install will take several minutes

If you're interested in this, we would be very happy to develop/test it and submit it as a PR!

begelundmuller:
I'm happy to hear that. We'd be happy to submit a PR for you to test and will try to prioritise it within a few weeks.

begelundmuller:
This has now been merged: https://github.com/marcboeker/go-duckdb/pull/40

The PR also includes pre-compiled static libraries inspired by https://github.com/rogchap/v8go. This means that on supported platforms, you get static binaries without having to wait for DuckDB to compile.

LasseJacobs:
@begelundmuller could you explain why having a single source file makes it easier to statically link? I thought this would not matter since in both cases (single vs multiple source files) you end up with a single archive file (.a)? But indeed I am unable to statically link the `libduckdb_static.a` file duckdb produces when I build it from source, while the archive file produced from the single file links statically without a problem.

I would like to understand why this is the case.

begelundmuller:
Hey @LasseJacobs, you're right that simply having one source file instead of multiple source files does not itself make a difference to our ability to compile from source or statically link an archive. 

[...]

PR #40, opened 2022-08-25
Use DuckDB amalgamation instead of shared libraries

begelundmuller:
See issue #38 for context.

This is a minimal PR that uses the DuckDB amalgamation to compile from source instead of relying upon a shared library. A few notes:

- No environment variables or flags required to test/run/build.
- This should work on any platform that supports CGO (incl. Windows).
- On first build, it compiles DuckDB, which is excruciatingly slow (~10 mins). Go caches the build on subsequent runs.
- It's necessary to check the amalgamation into the repository (like in `go-sqlite3`).

Next steps:

- Test on Linux and Windows.
- Double-check the compiler and linker flags set in `duckdb.go`.
- Add build flag to optionally link to `libduckdb` instead of compiling from source.
- What do you think of pre-compiling and checking in static libraries ('.a', not '.so') for common platforms? That's what [`rogchap/v8go`](https://github.com/rogchap/v8go) does (see [`deps`](https://github.com/rogchap/v8go/tree/master/deps) folder). That would still get rid of the build flags and produce a static binary, but you wouldn't have to wait for DuckDB to compile from source on most systems.


marcboeker:
Thanks for the PR. I've tested it successfully under macOS with a build time of around 6 minutes. It's really cool to get rid of external dependencies.

> Test on Linux and Windows.

I've tried to compile it on a MacBook Pro 2014 running Windows through Bootcamp. But after 12 hours it was still not done. The CPU usage on this Mac was around 20%. I've tried to increase the number of cores via `-j4` but this didn't work either. Have you had success compiling it under Windows in a reasonable time?
Maybe we could add a flag to make use of multiple cores while compiling?

> Double-check the compiler and linker flags set in duckdb.go.

What is linking against `ws2_32.dll` needed for?

> Add build flag to optionally link to libduckdb instead of compiling from source.

This is a great suggestion, especially if you want to reduce the initial build time or cross compile for another architecture without having to install several different cross compile chains.

> What do you think of pre-compiling and checking in static libraries ('.a', not '.so') for common platforms? That's what [rogchap/v8go](https://github.com/rogchap/v8go) does (see [deps](https://github.com/rogchap/v8go/tree/master/deps) folder). That would still get rid of the build flags and produce a static binary, but you wouldn't have to wait for DuckDB to compile from source on most systems.

Would it make sense to just download the different DuckDB libs and check them in? Or do you want to compile them separately on your own? 

One thing I have been thinking about is, that we should explain how to cross compile for other platforms and OSes. I've tried to use a cross compile chain on my M1 Mac for a 64bit linux, but `GOARCH=amd64 GOOS=linux CGO_ENABLED=1 CC=x86_64-unknown-linux-gnu-gcc go build -o linux examples/simple.go` failed with the error `ld: symbol(s) not found for architecture arm64` even though I've explicitly selected `amd64`.

I think your PR is an awesome addition to `go-duckdb` as it simplifies a lot. Hopefully we can merge this soon.

marcboeker:
Thanks for the additional build tags, the fix for the return of an empty row and your answers to my question. So far I've reviewed and tested it on macOS and it looks good to me. I like the new flexibility derived by the build tags a lot.

The pipeline has some hiccups with rules which it thinks are missing in the Makefile, but are there. Do you have any idea why?

We can ask the DuckDB creators to review the compiler and linker flags and then merge the PR. In my opinion the missing building blocks for Windows can be added after the merge, so that we can get the release out in the hands of testers quickly. What do you think?

begelundmuller:
Hey, a few comments:

- I tidied up some gcc compiler flags (I had originally copied them from v8go, but they don't seem to do anything important at this stage, and are not found in the sqlite3 project). Every remaining flag is necessary to make the build work – except for the `-O3` optimizer flag, which appears to be DuckDB's recommended optimization level for production (as far as I can tell, Go doesn't itself set optimization flags for CGo code – but I might be wrong).
- I think the main risks right now is that static linking might not be perfectly configured. DuckDB is getting statically linked, but the need to link in `libc++` in `cgo_static.go` doesn't feel quite right to me. But I don't think this should be a blocker, everything seems to work – we can fix if anyone encounters a problem with this approach.
- I updated the Github workflow to build and push the static libraries on any push to a non-master branch that changes the DuckDB amalgamation files. It failed before because the old approach couldn't correctly deal with cross-repo PRs (where the code is in another repo than the repo where the PR was filed). To show that the builds do work, [here](https://github.com/begelundmuller/go-duckdb/actions/runs/3135474063)'s a run that succeeded in building and pushing fresh static libs. The new workflow rule is:
```
push:
    branches-ignore:
      - master
    paths:
      - "duckdb.cpp"
      - "duckdb.h"
      - "duckdb.hpp"
```

By the way, I found this HN thread where someone requested a static DuckDB build for Go: https://news.ycombinator.com/item?id=31531219 😄

marcboeker:
Hi @begelundmuller 

Thanks for the adjustments. I've tested the latest version on macOS successfully. Are you fine with merging the current state so that further development is based on the new static builds?

That HN thread is very interesting. Great to see, that there is now a solution to this problem.

begelundmuller:
Yeah, it'd be great to merge this initial functionality!

marcboeker:
Awesome, it's merged now. Thank you for this great contribution, which makes using this driver much easier!

begelundmuller:
Nice – and thank you for building and maintaining this useful driver!

PR #45, opened 2022-10-09
Refactor binding

marcboeker:


marcboeker:
Hi @begelundmuller 
I've done some major refactoring to get rid of the query string interpolation for params and make use of the DuckDB bind API instead. This should make the driver more robust and use a common procedure for normal and prepared statements.

Would it be possible that you could check your code base against this branch to make sure, nothing is broken? That would be great. And if you have any remarks feel free to add comments. Thanks in advance!

statement.go
```diff
@@ -48,6 +48,10 @@ func (s *stmt) start(args []driver.Value) error {
 
 	for i, v := range args {
 		switch v := v.(type) {
+		case bool:
+			if rv := C.duckdb_bind_boolean(*s.stmt, C.idx_t(i+1), true); rv == C.DuckDBError {
```
begelundmuller, line 52: Shouldn't this bind to v?

begelundmuller:
Hi @marcboeker , I ran our tests against your PR and everything seems to work as expected. I also did a superficial read of the code in the PR and it looks great (left one small comment).

This is very useful, I actually just recently had to work around an arg binding bug that I wanted to get back to and investigate. This PR solves that bug!

marcboeker:
Hi @begelundmuller great to hear, that nothing has been broken and the PR unintentionally fixed a binding problem 🙂 Thanks for the review and the comment. I've fixed it, added a testcase and will merge it now.

PR #49, opened 2022-10-15
Support for all DuckDB data types and data type names

begelundmuller:
This PR is mainly about adding support for logical type names (like maps and structs) in `ColumnTypeDatabaseTypeName`. While going through the C API, I also noticed a few esoteric data types in DuckDB that were not covered in `scan` and added support for those (unsigned integers, `TIME`, and `TIMESTAMP_(NS|MS|S)`).

To make sure I didn't miss anything, I also reordered all switch statements to follow the ordering of [`enum DUCKDB_TYPE`](https://github.com/duckdb/duckdb/blob/master/src/include/duckdb.h#L66).

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -439,8 +439,10 @@ func typeName(t C.duckdb_type) string {
 		return "UUID"
 	case C.DUCKDB_TYPE_JSON:
 		return "JSON"
+	default:
+		// Should never happen
+		return ""
 	}
-	return ""
 }
 
 func logicalTypeName(lt C.duckdb_logical_type) string {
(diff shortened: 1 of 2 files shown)


rows.go
```diff
@@ -102,6 +102,8 @@ func scan(vector C.duckdb_vector, rowIdx C.idx_t) (any, error) {
 	}
 
 	ty := C.duckdb_vector_get_column_type(vector)
+	defer C.duckdb_destroy_logical_type(&ty)
```
marcboeker, line 105: Good catch, thank you! 🙂

Issue #50, opened 2022-10-15
Use math/big for HugeInt and time.Duration for interval


begelundmuller:
I'm running into some issues when serializing HugeInts. When I looked at the type, I also saw that `INTERVAL`s are returned as DuckDB types. Are you interested in a PR from me that moves to types from the standard library? I would replace `HugeInt` with `math/big.Int` and `Interval` with `time.Duration`.

marcboeker:
That would be great. The more we can move to the std lib the cleaner it will be. Thanks in advance.

begelundmuller:
Update here: https://github.com/marcboeker/go-duckdb/pull/52

I decided not to use `time.Duration` for intervals since intervals have a notion of days and months, which cannot be expressed with `time.Duration`.

PR #52, opened 2022-10-17
Use math/big.Int for HUGEINT

begelundmuller:
Solves https://github.com/marcboeker/go-duckdb/issues/50.

During this work, I realized we're using `float64` for `DECIMAL` types. Before merging this, should I extend the PR to use `big.Float` for `DECIMAL` types as well?

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -8,6 +8,7 @@ import "C"
 import (
 	"database/sql/driver"
 	"errors"
+	"math/big"
 	"unsafe"
 )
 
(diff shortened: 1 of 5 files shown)


marcboeker:
Thanks for the PR! At first I was a bit confused why big.Int was declared as pointer, but after reading the docs it makes sense. I would suggest to start a new PR for the DECIMAL type. Would you use math/big.Float for this?

PR #61, opened 2022-12-21
conn pool consistency fix

pjain1:
Fixes a consistency bug where changes made in one DuckDB connection are not immediately reflected in other DuckDB connections. Looks like the problem is that In the [go-duckdb](https://github.com/marcboeker/go-duckdb/blob/master/duckdb.go) driver, the Open function creates a new DB handle (using duckdb_open) and new connection handle (using duckdb_connect) each time. So the use of multiple DB handles seems to be the problem here. Thanks @begelundmuller for the analysis.

Driver now implements `DriverContext` so `sql.DB` will call `OpenConnector` to obtain a Connector and then invoke
that Connector's Connect method to obtain each needed connection instead of invoking Driver's Open method which was creating new `C.duckdb_database` each time. Now we create database object once on call to `OpenConnector` and call `C.duckdb_connect` on same database to obtain new connections on call to `Connect`.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -13,7 +13,6 @@ import (
 )
 
 type conn struct {
-	db     *C.duckdb_database
 	con    *C.duckdb_connection
 	closed bool
 	tx     bool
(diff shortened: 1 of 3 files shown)


marcboeker:
LGTM and thanks for the PR. I've never had this use case so far but I'm glad you have it, so connection pooling is now possible 🙂

pjain1:
Thanks for the quick review and merge.

PR #65, opened 2022-12-27
connection init function

### LINKED ISSUE #62: Support extensions when using a connection pool
We want to use the built-in connection pool in `database/sql` to run concurrent DuckDB queries. So far, we have not done so (disabled by setting `db.SetMaxOpenConns(1)`) because of two consistency problems:

1. Changes made in one connection were not immediately available in other connections – that should be fixed by https://github.com/marcboeker/go-duckdb/pull/61
2. Extensions need to be loaded on each new connection (they're not shared), but `database/sql` automatically creates new connections (and to my knowledge, there's no hook to do custom init)

We would like to submit a PR that addresses point 2. I can think of a two different solutions:

1. Support a custom `?extensions=json,parquet,...` config syntax, which the driver intercepts. It will then call `INSTALL` for each specified extension when the DB is opened, and subsequently call `LOAD` for each extension when a new connection is created.
2. Add an init callback option to the connector, which you can pass using [`sql.OpenDB`](https://pkg.go.dev/database/sql#OpenDB) – e.g. something like `sql.OpenDB(duckdb.NewConnector(dsn, connInitFn))` where `connInitFn` will be invoked each time a new connection is created for the DB handle.

What do you think of these options – or do you have a better idea for how we can solve this problem?

As an aside, we have done some benchmarking of using multiple DuckDB connections, and achieved a 2x performance boost for  processing 1k queries when using a connection pool of 10 (matching the number of cores on the system) instead of 1.

pjain1:
Fixes #62 

The driver implements `Execer` interface which is provided in the init callback function to perform any initializations. Another option is to provide `driver.Conn` handle but it has `Prepare(query string) (Stmt, error)` , `Close()` and `Begin()` methods. So users will have to use `Prepare` and then call `Exec` on `Stmt` interface to execute init queries. 

One more thing to note is that even `Execer` is deprecated, the new interface is `ExecerContext` but driver does not implement it yet.

### DIFF
--- a/duckdb_test.go
+++ b/duckdb_test.go
@@ -59,7 +59,9 @@ func TestConnPool(t *testing.T) {
 
 	// Get two separate connections and check they're consistent
 	conn1, err := db.Conn(context.Background())
+	require.NoError(t, err)
 	conn2, err := db.Conn(context.Background())
+	require.NoError(t, err)
 
 	res, err := conn1.ExecContext(context.Background(), "INSERT INTO foo VALUES ('lala', ?), ('lalo', ?)", 12345, 1234)
 	require.NoError(t, err)
@@ -107,10 +109,11 @@ func TestConnInit(t *testing.T) {
 
 	// Get two separate connections and check they're consistent
 	conn1, err := db.Conn(context.Background())
+	require.NoError(t, err)
 	conn2, err := db.Conn(context.Background())
+	require.NoError(t, err)
 
 	res, err := conn1.ExecContext(context.Background(), "CREATE TABLE example (j JSON)")
-	fmt.Println(err)
 	require.NoError(t, err)
 	ra, err := res.RowsAffected()
 	require.NoError(t, err)
--- a/duckdb_test.go
+++ b/duckdb_test.go
@@ -3,6 +3,7 @@ package duckdb
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


duckdb_test.go
```diff
@@ -84,6 +85,61 @@ func TestConnPool(t *testing.T) {
...
+		}
+		return nil
+	})
+	require.NoError(t, err)
+	db := sql.OpenDB(connector)
+	db.SetMaxOpenConns(2) // set connection pool size greater than 1
+	defer db.Close()
+
+	// Get two separate connections and check they're consistent
+	conn1, err := db.Conn(context.Background())
```
marcboeker, line 109: Maybe it makes sense to check for an error with `require.NoError(t, err)`

duckdb_test.go
```diff
@@ -84,6 +85,61 @@ func TestConnPool(t *testing.T) {
...
+	conn2, err := db.Conn(context.Background())
+
+	res, err := conn1.ExecContext(context.Background(), "CREATE TABLE example (j JSON)")
+	fmt.Println(err)
```
marcboeker, line 113: Could you please remove this?

marcboeker:
Thanks for the PR. LGTM. Maybe we should add some documentation, so that other understand how they can use the connection init callback?

duckdb_test.go
pjain1: done

duckdb_test.go
pjain1: removed

pjain1:
@marcboeker added some documentation, let me know if it makes sense. Thanks

marcboeker:
Awesome, thank you for this contribution.

PR #66, opened 2023-01-11
Support immediate query cancellation

### LINKED ISSUE #60: Support query cancellation and deadlines using contexts
Currently, the `go-duckdb` driver doesn't support cancelling queries when a context is cancelled. I just discovered it might be easy to add:

- By implementing [`database/sql/driver.StmtExecContext`](https://pkg.go.dev/database/sql/driver#StmtExecContext) instead of [`database/sql/driver.Stmt`](https://pkg.go.dev/database/sql/driver#Stmt), we get access to the query context and can check for cancellation.
- In [`statement.go`](https://github.com/marcboeker/go-duckdb/blob/72b078cd343c9eb3b457b279ab00838fd8ac3ce3/statement.go#L151), `duckdb_execute_prepared` is used to run a query. That will block for the duration of the entire query, which means we can't cancel a DuckDB query if the context gets cancelled. However, DuckDB provides an alternative that allows the caller to periodically regain control during query execution – it is `duckdb_pending_prepared` and related functions ([docs](https://duckdb.org/docs/api/c/api#duckdb_pending_prepared)). 

It looks like the Julia driver moved to using this ([PR](https://github.com/duckdb/duckdb/pull/4332)) to allow garbage collection to run. We could use it to periodically check the context, and cancel a query if the context has been cancelled or exceeded its deadline.

What do you think of this idea? Would be happy to submit a PR for this

pjain1:
Fixes #60 and #43 

Mainly query is executed in steps and context is checked before executing each step. It uses Pending Result Interface C APIs to achieve this. Reference - https://duckdb.org/docs/api/c/api#pending-result-interface

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -6,6 +6,7 @@ package duckdb
 import "C"
 
 import (
+	"context"
 	"database/sql/driver"
 	"errors"
 	"math/big"
@@ -26,6 +27,7 @@ func (c *conn) CheckNamedValue(nv *driver.NamedValue) error {
 	return driver.ErrSkip
 }
 
+// Deprecated: Use ExecContext instead. This method is not used in this project anywhere.
 func (c *conn) Exec(cmd string, args []driver.Value) (driver.Result, error) {
 	if c.closed {
 		panic("database/sql/driver: misuse of duckdb driver: Exec after Close")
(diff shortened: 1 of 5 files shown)


marcboeker:
Thanks for adding context support 🙂

PR #72, opened 2023-02-14
Update DuckDB to 0.7.0

marcboeker:


begelundmuller:
Hey, I'm eager to try this out! I noticed the tests fails, it looks like it's just due to DuckDB removing the `DUCKDB_TYPE_JSON` type. Let me know if there's any way I can help.

marcboeker:
I was a bit confused about that and already [asked the DuckDB](https://github.com/duckdb/duckdb/discussions/6288#discussioncomment-4979914) devs about this.

I've already removed the DuckDB JSON type, but now I'm having problems with the `MAP` type in the tests. I have to further investigate. The current state is pushed, if you have any ideas.

begelundmuller:
Bummer – maybe it's related to: https://github.com/duckdb/duckdb/pull/5768?

marcboeker:
Yeah that could exactly be the breaking change. I've [asked them](https://github.com/duckdb/duckdb/discussions/6288#discussioncomment-4986991) about this.

A lot of breaking changes for a minor version upgrade. To be fair, it's still pre 1.0.0 😃

marcboeker:
@begelundmuller I've [pushed a fix](https://github.com/marcboeker/go-duckdb/pull/72/commits/ce83577f46008aef590715aa39170b1a306a4f56) for the new map layout. Could you please have a look at it, if this looks good to you. Thanks!

begelundmuller:
@marcboeker, took a look at the changes and they seem good to me.

I'm wondering if it would be more performant to scan into a `[]struct {Key any, Value any}`, but that would require refactoring the struct scanning, so doesn't make sense to do here (and I'm guessing Go's `map` is pretty performant anyway).

marcboeker:
> took a look at the changes and they seem good to me.

Thanks!

> I'm wondering if it would be more performant to scan into a `[]struct {Key any, Value any}`, but that would require refactoring the struct scanning, so doesn't make sense to do here (and I'm guessing Go's `map` is pretty optimized anyway).

This would definitely be a better way. I was hoping that having a `map` with only two keys should be fast enough. But we can keep this improvement as a separate ticket. I'll create one.

PR #74, opened 2023-02-19
Return err instead of panic for invalid map keys

begelundmuller:
It seems DuckDB supports more map key types than Go, causing the assignment `out[key] = val` to panic. This PR adds a check that map key types are comparable and returns an error if not.

An alternative would be to not scan DuckDB maps into Go `map`, but instead directly return the list object containing `{"key": ..., "value": ...}` maps. What do you prefer?

### DIFF
--- a/duckdb_test.go
+++ b/duckdb_test.go
@@ -286,6 +286,12 @@ func TestMap(t *testing.T) {
 		}, m)
 	})
 
+	t.Run("select nested map", func(t *testing.T) {
+		var m Map
+		err := db.QueryRow("SELECT map([map([1], [1]), map([2], [2])], ['a', 'e'])").Scan(&m)
+		require.ErrorIs(t, err, errUnsupportedMapKeyType)
+	})
+
 	t.Run("insert map", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


marcboeker:
Good catch, thanks for the fix! I suggest we keep it as it is at the moment and if there is demand for non comparable map keys we can switch to the list approach.

begelundmuller:
Sounds good – thanks for the fast review!

begelundmuller:
Just adding a note that I discovered that the Python driver takes the list approach:

```py
>>> duckdb.sql("SELECT map(['B', 'A'], [map([1, 2], [3,4]), map([1, 2], [3,4])]);").fetchall()
[({'key': ['B', 'A'], 'value': [{'key': [1, 2], 'value': [3, 4]}, {'key': [1, 2], 'value': [3, 4]}]},)]
```

Might be worth doing at some point to support all cases, even though it has worse ergonomics.

marcboeker:
This looks so weird to me, but yeah it would make sense to be more flexible. Shouldn't be that complicated to implement though.

PR #76, opened 2023-02-21
Add duckdb open ext

rakeshsharma14317:
Adding use of duckdb_open_ext for returning error msg for db file lock and version upgrade errors


marcboeker:
Thanks for the PR!

PR #79, opened 2023-03-03
Upgrade to  DuckDB 0.7.1

begelundmuller:
I noticed that DuckDB 0.7.1 also added struct field name escaping in type names. I added a fix to reflect the change.

Here's an example of how DuckDB 0.7.1 escapes struct field names in type names:
```
D create or replace table x as select {'key1 (,) \ \" "" "': 1} as col;
D select data_type from information_schema.columns;
┌──────────────────────────────────────────┐
│                data_type                 │
│                 varchar                  │
├──────────────────────────────────────────┤
│ STRUCT("key1 (,) \ \"" """" """ INTEGER) │
└──────────────────────────────────────────┘
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DUCKDB_VERSION=0.7.0
+DUCKDB_VERSION=0.7.1
 
 .PHONY: install
 install:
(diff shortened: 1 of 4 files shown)


marcboeker:
Thanks and have a nice weekend!

PR #83, opened 2023-03-14
fix: release allocated memory for data chunks

### LINKED ISSUE #81: Possible memory leak when processing parquet files
We have seen that the used memory keeps increasing when we are processing parquet files with the go-duckdb library.
After we have processed a single parquet file and closes the db connection I assume that the used memory is released. However, as seen in the example graph below that is not the case, and for repetitive processing of the same source the memory continues to increase.

## Expected behaviour
The memory used for querying a parquet file is released when it is finished.

## Actual behaviour
The memory continues to increase and it does not seem to ever release it.
image

## Reproduce
I have created a tiny project to reproduce the issue that can be found here: https://github.com/goober/duckdb-memory-leak

goober:
This PR aims to fix #81  by making sure that the memory allocated to each chunk of data from the result will be released before next chunk is fetched. 

## Memory usage without the fix
The memory keeps increasing each time the parquet-file is processed.

The graph below shows the memory usage when running inside a docker container
memory usage without fix

The video below shows the memory usage when running without docker
https://user-images.githubusercontent.com/124244/224948192-fd569a41-0140-4516-aaec-4c80f4f9fcba.mov

## Memory usage with the fix
When running inside a docker container you can see that the memory stabilizes and will not increase further regardless of recurrent file processes.
memory usage with fix

When we are running without docker you can see that the the memory is released and does not increase on recurrent requests
https://user-images.githubusercontent.com/124244/224948151-72abef6d-5974-462a-8830-c3bcd41c44a7.mov

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -56,6 +56,7 @@ func (r *rows) Columns() []string {
 
 func (r *rows) Next(dst []driver.Value) error {
 	for r.chunkRowIdx == r.chunkRowCount {
+		C.duckdb_destroy_data_chunk(&r.chunk)
 		if r.chunkIdx == r.chunkCount {
 			return io.EOF
 		}
@@ -257,6 +258,7 @@ func (r *rows) ColumnTypeDatabaseTypeName(index int) string {
 }
 
 func (r *rows) Close() error {
+	C.duckdb_destroy_data_chunk(&r.chunk)
 	C.duckdb_destroy_result(&r.res)
 
 	if r.stmt != nil {


marcboeker:
@goober Awesome, thank you for debugging and fixing this.

rows.go
```diff
@@ -257,6 +258,7 @@ func (r *rows) ColumnTypeDatabaseTypeName(index int) string {
 }
 
 func (r *rows) Close() error {
+	C.duckdb_destroy_data_chunk(&r.chunk)
 	C.duckdb_destroy_result(&r.res)
```
chrisirhc, line 262: Is it necessary to set these as `nil`?

PR #88, opened 2023-04-24
Pass the full connection string to duckdb

elefeint:
Hello, and thank you for creating the driver -- it works great with all traditional uses of DuckDB on the filesystem.

This PR unlocks a new ability recently added in DuckDB to [detect and attach to different database types](https://github.com/duckdb/duckdb/blob/v0.7.1/src/main/database.cpp#L200-L205). Currently, the driver only passes the parsed out _path_ component from the user-provided connection string to duckdb itself. With this change, the full user-provided connection string up to the query will be passed through. 

I've run the automated tests and manually tested the driver against duckdb, motherduck and sqlite.

Since I am currently testing the Go data ecosystem integrations with duckdb (mostly from the perspective of MotherDuck, but it ends up exercising standalone duckdb nicely, too), if I find other edge cases -- would you prefer filed issues for everything, or only the pull requests for the ones I know how to solve?


elefeint:
Definitely; I'll add a test -- although sqlite would have worked even without this change, thanks to duckdb using magic bytes autodetection.

marcboeker:
Awesome, thank you!

elefeint:
@marcboeker sorry, would it be possible to cut a release with this change?

marcboeker:
@elefeint Sure, v1.2.3 has been released.

PR #90, opened 2023-05-05
feat: better decimal support

TN1ck:
Hey all,

go-duckdb currently offers rudimentary decimal support by casting them to float64. This is problematic as precision is lost during this conversion, so one cannot use this library currently for anything decimal involved that requires no loss of precision (e.g. money handling).

I tried adding  decimal support that transparently surfaces the decimals from duckdb, so one can do whatever you want with them.

As an example on how to use them with the popular decimal library from shopspring:

```go
import (
  "github.com/shopspring/decimal"
)

shopspringDecimal := decimal.NewFromBigInt(f.Value, -int32(f.Scale))
```

There is a bug when querying for decimal values like seen in the commented out tests. It's a bug with duckdbs query parser that appears when using the amalgamation. The following text was the initial debugging, but it's a problem that surfaces in duckdb code.

<details>
  <summary>Open initial investigation</summary>
  
It's _mostly_ working, but there is a strange bug that I seems to be deeper then what I am doing here.
Decimals that have hugeints are not correctly converted, but instead have the same precision what seems a float64 would have. I haven't really dig deep there, but as I'm using the already available `hugeIntToNative` that returns the wrong number. It must be somewhere there. The normal hugeint conversion works fine though and doesn't exhibit this problem.

Happy to look into this if anybody has some pointers.

### UPDATE:

I cross checked with the official JDBC driver to see where the issue might appear. First I tried to align the current hugeIntToNative to the [java version](https://github.com/duckdb/duckdb/blob/dd0e0da3f0d1dfc8d9cb0d7f02571b223c496c44/tools/jdbc/src/main/java/org/duckdb/DuckDBResultSet.java#L771):

```go
var ulongMultiplier, _ = new(big.Int).SetString("18446744073709551616", 10)
func hugeIntToNative(hi C.duckdb_hugeint) *big.Int {
	i := big.NewInt(int64(hi.upper))
	i.Mul(i, ulongMultiplier)
	i.Add(i, new(big.Int).SetUint64(uint64(hi.lower)))
	return i
}
```

But this yields the exact same results.

I went then looking how the `lower` and `upper` values look in the java vs go version. I added a new java test case to the jdbc tests.

<details>
  <summary>The java test</summary>

```java
	public static void test_decimal_again() throws Exception {
		Connection conn = DriverManager.getConnection("jdbc:duckdb:");
		Statement stmt = conn.createStatement();

		ResultSet rs = stmt.executeQuery(
				"SELECT 123456789.01234567890123456789 :: DECIMAL(38, 20) d");

		assertTrue(rs.next());
		BigDecimal result = rs.getBigDecimal("d");
		assertEquals(result, new BigDecimal("123456789.01234567890123456789"));

		assertFalse(rs.next());
		rs.close();
		stmt.close();
		conn.close();
	}
```
</details>

And looked what the values are. They are different:

Java yields:
* upper - `669260594`
* lower - `5097733592125636885 (0x46bec9b16e398115)`

Go yields:
* upper - `669260594`
* lower - `5097731730555535369 (0x46bec80000000009)`

Which looks like the `lower` value gets cut off somewhere. Most likely in the `get` function.
</details>

### DIFF
--- a/duckdb_test.go
+++ b/duckdb_test.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math"
 	"math/big"
 	"reflect"
 	"testing"
@@ -359,6 +358,12 @@ func TestList(t *testing.T) {
 	})
 }
 
+func compareDecimal(t *testing.T, want Decimal, got Decimal) {
+	require.Equal(t, want.Scale, got.Scale)
+	require.Equal(t, want.Width, got.Width)
+	require.True(t, want.Value.Cmp(got.Value) == 0)
+}
+
 func TestDecimal(t *testing.T) {
 	t.Parallel()
 	db := openDB(t)
@@ -366,9 +371,9 @@ func TestDecimal(t *testing.T) {
 
 	t.Run("decimal widths", func(t *testing.T) {
 		for i := 1; i <= 38; i++ {
-			var f float64 = 999
+			var f Decimal
 			require.NoError(t, db.QueryRow(fmt.Sprintf("SELECT 0::DECIMAL(%d, 1)", i)).Scan(&f))
-			require.Equal(t, float64(0), f)
+			require.Equal(t, Decimal{Width: uint8(i), Value: big.NewInt(0), Scale: 1}, f)
 		}
 	})
 
@@ -377,26 +382,44 @@ func TestDecimal(t *testing.T) {
 			(1.23 :: DECIMAL(3, 2)),
 			(123.45 :: DECIMAL(5, 2)),
 			(123456789.01 :: DECIMAL(11, 2)),
-			(1234567890123456789.234 :: DECIMAL(22, 3)),
+			(12345678901234.567 :: DECIMAL(22, 3)),
 		) v
 		ORDER BY v ASC`)
 		require.NoError(t, err)
 		defer rows.Close()
 
-		want := []float64{1.23, 123.45, 123456789.01, 1234567890123456789.234}
+		// TODO: We cannot go higher than this right due to an issue with duckdb's query parser.
+		bigNumber, success := new(big.Int).SetString("12345678901234567", 10)
+		require.True(t, success, "failed to parse big number")
+		want := []Decimal{
+			{Value: big.NewInt(1230), Width: 22, Scale: 3},
+			{Value: big.NewInt(123450), Width: 22, Scale: 3},
+			{Value: big.NewInt(123456789010), Width: 22, Scale: 3},
+			{Value: bigNumber, Width: 22, Scale: 3}}
 		i := 0
 		for rows.Next() {
-			var fs float64
+			var fs Decimal
 			require.NoError(t, rows.Scan(&fs))
-			require.Equal(t, want[i], fs)
+			compareDecimal(t, want[i], fs)
 			i++
 		}
 	})
 
+	// TODO: We get lower precision numbers here due to a bug it what seems duckdb's query parser.
+	// t.Run("huge decimal", func(t *testing.T) {
+	// 	bigNumber, success := new(big.Int).SetString("12345678901234567890123456789", 10)
+	// 	require.True(t, success, "failed to parse big number")
+	// 	var f Decimal
+	// 	require.NoError(t, db.QueryRow("SELECT 123456789.01234567890123456789 :: DECIMAL(38, 20)").Scan(&f))
+	// 	compareDecimal(t, Decimal{Value: bigNumber, Width: 38, Scale: 20}, f)
+	// })
+
 	t.Run("huge decimal", func(t *testing.T) {
-		var f float64
-		require.NoError(t, db.QueryRow("SELECT 123456789.01234567890123456789 :: DECIMAL(38, 20)").Scan(&f))
-		require.True(t, math.Abs(float64(123456789.01234567890123456789)-f) < 0.0000001)
+		bigNumber, success := new(big.Int).SetString("123456789012345612", 10)
+		require.True(t, success, "failed to parse big number")
+		var f Decimal
+		require.NoError(t, db.QueryRow("SELECT 1234567890123456.12 :: DECIMAL(38, 2)").Scan(&f))
+		compareDecimal(t, Decimal{Value: bigNumber, Width: 38, Scale: 2}, f)
 	})
 }
 
@@ -844,9 +867,9 @@ func TestTypeNamesAndScanTypes(t *testing.T) {
 		},
 		// DUCKDB_TYPE_DECIMAL
 		{
-			sql:      "SELECT 31::DECIMAL(30,20) AS col",
-			value:    float64(31),
-			typeName: "DECIMAL(30,20)",
+			sql:      "SELECT 31::DECIMAL(30,17) AS col",
+			value:    Decimal{Value: big.NewInt(3100000000000000000), Width: 30, Scale: 17},
+			typeName: "DECIMAL(30,17)",
 		},
 		// DUCKDB_TYPE_TIMESTAMP_S
 		{
--- a/rows.go
+++ b/rows.go
@@ -214,7 +214,7 @@ func (r *rows) ColumnTypeScanType(index int) reflect.Type {
 	case C.DUCKDB_TYPE_BLOB:
 		return reflect.TypeOf([]byte{})
 	case C.DUCKDB_TYPE_DECIMAL:
-		return reflect.TypeOf(float64(0))
+		return reflect.TypeOf(Decimal{})
 	case C.DUCKDB_TYPE_TIMESTAMP_S:
 		return reflect.TypeOf(time.Time{})
 	case C.DUCKDB_TYPE_TIMESTAMP_MS:
@@ -360,7 +360,13 @@ func scanStruct(ty C.duckdb_logical_type, vector C.duckdb_vector, rowIdx C.idx_t
 	return data, nil
 }
 
-func scanDecimal(ty C.duckdb_logical_type, vector C.duckdb_vector, rowIdx C.idx_t) (float64, error) {
+type Decimal struct {
+	Width uint8
+	Scale uint8
+	Value *big.Int
+}
+
+func scanDecimal(ty C.duckdb_logical_type, vector C.duckdb_vector, rowIdx C.idx_t) (Decimal, error) {
 	scale := C.duckdb_decimal_scale(ty)
 	width := C.duckdb_decimal_width(ty)
 	var value C.duckdb_hugeint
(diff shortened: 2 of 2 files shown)


marcboeker:
Hi @TN1ck 

thank you for the PR which is a great addition. I've put a lot of time into debugging the problem with the lower part of the `big.Int` value. 

After some tests I was running this query from a test Go app, where I was trying to scan a decimal into a `big.Int` by accident:

```
res := db.QueryRow("SELECT 1234567890123456789.234::DECIMAL(22, 3)")
var bi *big.Int
res.Scan(&bi)
```

And the error from DuckDB was

```
panic: sql: Scan error on column index 0, name "CAST(1234567890123456774.148 AS DECIMAL(22,3))": unsupported Scan, storing driver.Value type duckdb.Decimal into type *big.Int
```

What confuses me is, that the value in the `CAST` (`1234567890123456774.148 AS DECIMAL(22,3)`) differs from the provided value (`1234567890123456789.234::DECIMAL(22, 3)`) and is exactly the value, that a `Scan()` into the `big.Int` of the `Decimal.Value` returns when I change my type to `Decimal`.
I don't think that the value is truncated in the `get` but instead that DuckDB internally mishandles the provided value from the query.

To prove that, I have created a DuckDB database manually with the following row:

```
D create table test(val decimal(22,3));
D insert into test values(1234567890123456789.234::DECIMAL(22, 3));
D select * from test;
┌─────────────────────────┐
│           val           │
│      decimal(22,3)      │
├─────────────────────────┤
│ 1234567890123456789.234 │
└─────────────────────────┘
```

Then I have loaded and queried the value and received the following response `Decimal{22 3 1234567890123456789234}`, which is correct.

As we simply provide the following query `SELECT 1234567890123456789.234::DECIMAL(22, 3)` as a string to DuckDB, I think the problem lies in their SQL parser or somewhere deeper.

Apart from this, I have refactored the decimal test. Maybe you could incorporate this into your PR:
https://github.com/marcboeker/go-duckdb/blob/3d852bfd69c7bfc9ee386bd86a41d92a7e7b73b2/duckdb_test.go#L367-L407

I would suggest to report this issue to the DuckDB developers as it seems to be a problem on their side.

Maybe @begelundmuller could also review this PR but for me it looks good.
Thanks!

TN1ck:
Awesome, great for digging into this @marcboeker. Wouldn't have thought that it's a bug of duckdb, but it looks like your're right. Looked forever at it myself and coudn't find anything. 
Thanks for refactoring those tests, I'll use your changes.

But in this state, the tests are not passing.  I tried to change the syntax e.g. `SELECT '123456789.01234567890123456789' :: TEXT :: DECIMAL(29, 20)`, but it always returns the wrong value, duckdb must do something wrong when parsing those Decimals.

Should we disable these tests for now? Or use the wrong values?

Also what do you think about also providing the float64 value, so users who rely on it have an easy time?

```go
type Decimal struct {
	Width          uint8
	Scale          uint8
	Value          *big.Int
	InexactFloat64 float64
}

func scanDecimal(ty C.duckdb_logical_type, vector C.duckdb_vector, rowIdx C.idx_t) (Decimal, error) {
	scale := C.duckdb_decimal_scale(ty)
	width := C.duckdb_decimal_width(ty)
	var value C.duckdb_hugeint
	switch C.duckdb_decimal_internal_type(ty) {
	case C.DUCKDB_TYPE_SMALLINT:
		value.lower = C.uint64_t(get[uint16](vector, rowIdx))
	case C.DUCKDB_TYPE_INTEGER:
		value.lower = C.uint64_t(get[uint32](vector, rowIdx))
	case C.DUCKDB_TYPE_BIGINT:
		value.lower = C.uint64_t(get[uint64](vector, rowIdx))
	case C.DUCKDB_TYPE_HUGEINT:
		i := get[C.duckdb_hugeint](vector, rowIdx)
		value.lower = i.lower
		value.upper = i.upper
	default:
		return Decimal{}, errInvalidType
	}

	nativeValue := hugeIntToNative(value)
	if nativeValue == nil {
		return Decimal{}, fmt.Errorf("unable to convert hugeint to native type")
	}

	decimal := C.duckdb_decimal{width, scale, value}
	return Decimal{
		Width:          uint8(width),
		Scale:          uint8(scale),
		Value:          nativeValue,
		InexactFloat64: float64(C.duckdb_decimal_to_double(decimal)),
	}, nil
}
```

I think that would be the good way without having too much decimal logic in this library. It leads to more memory use, as now we also save the float64.

TN1ck:
Looked into it some more and the bug happens in the cpp code, in `duckdb_translate_result`. I logged some values here https://github.com/TN1ck/go-duckdb/pull/1, go sends everything correctly and then duckdb suddenly has a wrong value.

marcboeker:
Thanks for your reply @TN1ck. I would suggest to disable the broken tests and make a comment in the code so that we can track this and re-enable them, once DuckDB has fixed the issue on their side.

Your proposed change with the `Decimal` would change the public API of the current major version `1.X.X`. So it would make sense, to increase the major version to `2.X.X` to not confuse people with a broken API. 
As the `Decimal` feature works on our side, I would skip the `InexactFloat64` field and stick to the plain `Decimal` implementation. People who are already using the `DECIMAL` type in DuckDB trough go-duckdb, need to migrate their code anyways.

@begelundmuller Were you currently using the decimals trough go-duckdb? What would be you desired behaviour regarding a major version increase?

TN1ck:
> Thanks for your reply @TN1ck. I would suggest to disable the broken tests and make a comment in the code so that we can track this and re-enable them, once DuckDB has fixed the issue on their side.
> 
> Your proposed change with the `Decimal` would change the public API of the current major version `1.X.X`. So it would make sense, to increase the major version to `2.X.X` to not confuse people with a broken API. As the `Decimal` feature works on our side, I would skip the `InexactFloat64` field and stick to the plain `Decimal` implementation. People who are already using the `DECIMAL` type in DuckDB trough go-duckdb, need to migrate their code anyways.
> 
> @begelundmuller Were you currently using the decimals trough go-duckdb? What would be you desired behaviour regarding a major version increase?

Sounds good, dropped the InexectFloat stuff and commented out the tests and changed the existing ones to pass.

I'm still debugging what is wrong with duckdb, the bug does not occur when using their unit tests, but creating an amalgation from HEAD makes it show up here. So I think it's something with the Amalgation process.

marcboeker:
@TN1ck I would also suggest to compare BigInts like this: https://github.com/marcboeker/go-duckdb/blob/3d852bfd69c7bfc9ee386bd86a41d92a7e7b73b2/duckdb_test.go#LL364C1-L365C1

Have you already had the opportunity to report your findings to the DuckDB maintainers?

If your PR is merged, it would be great to inform people about incorrect values, which occur ATM and point them to the bug report.

TN1ck:
> @TN1ck I would also suggest to compare BigInts like this: https://github.com/marcboeker/go-duckdb/blob/3d852bfd69c7bfc9ee386bd86a41d92a7e7b73b2/duckdb_test.go#LL364C1-L365C1

Will do the change! Personally I like the error comparison using strings as it makes the error messages easier to read.

> 
> Have you already had the opportunity to report your findings to the DuckDB maintainers?
> 

Yes, but only in their discord for now. I'll make a real issue today. The problem lies in the Amalgamation build, it doesn't succeed in their own test suit. For their own drivers they are moving away from it (e.g. their NodeJS one). So it might be worthwhile to look into building duckdb not from the amalgamation files, I think they'll deprecate it eventually, it doesn't seem that maintained. See https://github.com/TN1ck/duckdb/pull/1

> If your PR is merged, it would be great to inform people about incorrect values, which occur ATM and point them to the bug report.

I make an issue in their repo and link it in the code, makes it easier for referencing. 


> @begelundmuller Thanks for the review. I would suggest to increase the major version, as we're introducing an API change. Even if it is a small one.

TN1ck:
Created an issue at https://github.com/duckdb/duckdb/issues/7539

marcboeker:
@TN1ck Thanks for creating the issue @duckdb. I'm not quite sure if we should merge this PR now, with the potential problem of false results or if we should wait until the amalgamation file issue has been fixed. An alternative would be to concatenate our own amalgamation file with your potential fix to be independent from DuckDB's response to your issue. What do you think?

TN1ck:
> @TN1ck Thanks for creating the issue @duckdb. I'm not quite sure if we should merge this PR now, with the potential problem of false results or if we should wait until the amalgamation file issue has been fixed. An alternative would be to concatenate our own amalgamation file with your potential fix to be independent from DuckDB's response to your issue. What do you think?

No hurry from side about merging this, so we can totally wait. But I haven't yet found a fix sadly for the issue, I merely verified that their own tests fail against the amalgamation build (also verified that the decimal issue is due to that). I'll continue to work on it to speed it up, but can't promise anything.

TN1ck:
> > @TN1ck Thanks for creating the issue @duckdb. I'm not quite sure if we should merge this PR now, with the potential problem of false results or if we should wait until the amalgamation file issue has been fixed. An alternative would be to concatenate our own amalgamation file with your potential fix to be independent from DuckDB's response to your issue. What do you think?
> 
> No hurry from side about merging this, so we can totally wait. But I haven't yet found a fix sadly for the issue, I merely verified that their own tests fail against the amalgamation build (also verified that the decimal issue is due to that). I'll continue to work on it to speed it up, but can't promise anything.

I actually found the issue (wrong function selection due to overloading) and extended my bug with the investigation, they already have a fix out https://github.com/duckdb/duckdb/pull/7661, basically renaming the function. I'll check tonight if that fixes the issue and then one could either wait for the next duckdb release or recreate the amalgamation from head.

TN1ck:
Verified that this works now in https://github.com/TN1ck/go-duckdb/pull/2 🎉 

@marcboeker How should we proceed? Change this PR to include the Amalgamation files or do you want to wait until the next duckdb release is created (could take some time as it just released)?

marcboeker:
@TN1ck Great to hear that it works. How have you created the amalgamation file? Using `make amaldebug`? I've tried that with applying the patch for the `quantile.cpp` beforehand but received the following error after compiling the go-duckdb libs and running the tests:

```
make test
./duckdb.hpp:173:16: warning: defaulted function definitions are a C++11 extension [-Wc++11-extensions]
./duckdb.hpp:175:2: error: unknown type name 'constexpr'
...
```

I'd like to make sure that I'm also able to build the amalgamation file.
Thanks in advance!

TN1ck:
> @TN1ck Great to hear that it works. How have you created the amalgamation file? Using `make amaldebug`? I've tried that with applying the patch for the `quantile.cpp` beforehand but received the following error after compiling the go-duckdb libs and running the tests:
> 
> ```
> make test
> ./duckdb.hpp:173:16: warning: defaulted function definitions are a C++11 extension [-Wc++11-extensions]
> ./duckdb.hpp:175:2: error: unknown type name 'constexpr'
> ...
> ```
> 
> I'd like to make sure that I'm also able to build the amalgamation file. Thanks in advance!

Basically! But to create the amalgamation, you don't have to build it, just execute the first bit of the `make amaldebug` command: `python3 scripts/amalgamation.py`

That will create the files under `src/amalgamation`.  Then you can copy & paste them into this project and surround them by the `#ifdef GODUCKDB_FROM_SOURCE ... #endif`.

I'm not sure though why you can't execute `make amaldebug`, that definitely works for me locally with the current duckdb head. I didn't try to just apply the patch as it's now merged.

marcboeker:
Thanks...I missed the `#ifdef GODUCKDB_FROM_SOURCE ... #endif` step 🙈. Now it works and I was able to test it successfully. I noticed that the refactoring for the tests was missing in this PR, so I took your PR and added them back again. Please check #98 if everything looks good to you. After your 👍 we can merge #98 into master.

TN1ck:
> Thanks...I missed the `#ifdef GODUCKDB_FROM_SOURCE ... #endif` step 🙈. Now it works and I was able to test it successfully. I noticed that the refactoring for the tests was missing in this PR, so I took your PR and added them back again. Please check #98 if everything looks good to you. After your 👍 we can merge #98 into master.

Thank you, looks great! I close this PR then when your merge it. Funny how this ended up fixing a bug in duckdbs source, would have not expected this when I started 😄

marcboeker:
Thanks for your hard work and all the debugging! It's merged now. Great to have this huge improvement merged. Will bump the version to 2.0.0 shortly.

Issue #55, opened 2022-11-30
Use ExtractStatement to split query into multiple statements


marcboeker:
DuckDB offers a method in the C++ API to split a cmd string into multiple statements. Now the method has been ported to the C API in a first [PR](https://github.com/duckdb/duckdb/pull/5524). I've started to implement the the `duckdb_extracted_statements` function.

What should be the expected behaviour if the following code is given?

`db.Exec("INSERT INTO foo VALUES (?); INSERT INTO bar VALUES (?);", "hello", "world")`

a) Should we extract the single statements and distribute the parameters over the statements. The first statement should receive `hello` and the second `world`.

b) Should we provide all arguments to all statements. Statement 1 and 2 would then receive `hello` and `world` which would result in an error, as each statement only expects one argument.

c) Should we throw an error if multiple statements are provided along with one or more arguments like it is implemented at the moment?

My personal opinion would be `c` or maybe `a`. But `a` feels counterintuitive and contains too much magic. `b` seems totally weird to me.

Another thing would be, that only the last statement would return a result, which also seems confusing. Have you ever had the use-case with multiple `SELECT` statements in one query string in your application?

What do you think @begelundmuller ?

Thanks!

marcboeker:
Thanks for your detailed opinion. I agree with you, that having multiple SELECTs in one statement is kinda weird.

It seems, that the Python driver [applies the given parameters to each query](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/src/pyconnection.cpp#L183) instead of distributing them.
The C++ API seems to have the [identical behavior](https://github.com/duckdb/duckdb/blob/942d917145ae4a562b5e3d3649fdabe5ae202305/test/api/test_prepared_api.cpp#L395). This is similar to `b`.

And the Python driver also returns the [result of the last statement](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/src/pyconnection.cpp#L161).

I think I need to do more research on this to prevent total confusion.

begelundmuller:
If that's what the official drivers do, I would probably just do the same :)

PR #91, opened 2023-05-12
use extract statement api - better support for multiple statments

pjain1:
Fixes https://github.com/marcboeker/go-duckdb/issues/55

Use extract statement api to split query into multiple statements. Then execute all but the last statement (without applying any args), if there are no errors then apply the args (if any) to the last statement and return the results. 

This behaviour is same as [Python](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/src/pyconnection.cpp#L440) and [C](https://github.com/duckdb/duckdb/blob/master/test/api/capi/test_capi_extract.cpp#L6) driver.

I don't think [Prepare](https://github.com/marcboeker/go-duckdb/blob/master/connection.go#L69) method on `Conn` can be called with multiple statements, so it still uses older `prepareStmt` method without using extract api.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -91,6 +91,7 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	// prepare and execute last statement with args and return result
 	stmt, err := c.prepareExtractedStmt(stmts, size-1)
 	if err != nil {
+		stmt.Close()
 		return nil, err
 	}
 
--- a/connection.go
+++ b/connection.go
@@ -34,6 +34,7 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 	}
 
 	stmts, size, err := c.extractStmts(query)
+	defer C.duckdb_destroy_extracted(&stmts)
 	if err != nil {
 		return nil, err
 	}
@@ -46,10 +47,7 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		}
 		// send nil args to execute statement and ignore result
 		_, err = stmt.ExecContext(ctx, nil)
-		if err != nil {
-			return nil, err
-		}
-		err = stmt.Close()
+		stmt.Close()
 		if err != nil {
 			return nil, err
 		}
@@ -60,8 +58,6 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		return nil, err
 	}
 	defer stmt.Close()
-
-	C.duckdb_destroy_extracted(&stmts)
 	return stmt.ExecContext(ctx, args)
 }
 
@@ -71,6 +67,7 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	}
 
 	stmts, size, err := c.extractStmts(query)
+	defer C.duckdb_destroy_extracted(&stmts)
 	if err != nil {
 		return nil, err
 	}
@@ -83,14 +80,7 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 		}
 		// send nil args to execute statement and ignore result
 		rows, err := stmt.QueryContext(ctx, nil)
-		if err != nil {
-			return nil, err
-		}
-		err = rows.Close()
-		if err != nil {
-			return nil, err
-		}
-		err = stmt.Close()
+		rows.Close()
 		if err != nil {
 			return nil, err
 		}
@@ -101,7 +91,6 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 		return nil, err
 	}
 
-	C.duckdb_destroy_extracted(&stmts)
 	return stmt.QueryContext(ctx, args)
 }
 
@@ -173,14 +162,14 @@ func (c *conn) extractStmts(query string) (C.duckdb_extracted_statements, C.idx_
 	var stmts C.duckdb_extracted_statements
 	var stmtsCount C.idx_t
 	stmtsCount = C.duckdb_extract_statements(*c.con, cquery, &stmts)
-	err := C.GoString(C.duckdb_extract_statements_error(stmts))
 
-	if err != "" {
-		C.duckdb_destroy_extracted(&stmts)
-		return nil, 0, errors.New(err)
-	}
 	if stmtsCount == 0 {
+		err := C.GoString(C.duckdb_extract_statements_error(stmts))
 		C.duckdb_destroy_extracted(&stmts)
+
+		if err != "" {
+			return nil, 0, errors.New(err)
+		}
 		return nil, 0, errors.New("no statements found")
 	}
 
--- a/connection.go
+++ b/connection.go
@@ -33,9 +33,9 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		panic("database/sql/driver: misuse of duckdb driver: ExecContext after Close")
 	}
 
-	stmts, size := c.extractStmts(query)
-	if size == 0 {
-		return nil, errors.New("no statements found")
+	stmts, size, err := c.extractStmts(query)
+	if err != nil {
+		return nil, err
 	}
 
 	// execute all statements but the last one, apply args only to last statement and return result of last statement
@@ -44,9 +44,12 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		if err != nil {
 			return nil, err
 		}
-		defer stmt.Close()
 		// send nil args to execute statement and ignore result
 		_, err = stmt.ExecContext(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		err = stmt.Close()
 		if err != nil {
 			return nil, err
 		}
@@ -57,6 +60,8 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
 		return nil, err
 	}
 	defer stmt.Close()
+
+	C.duckdb_destroy_extracted(&stmts)
 	return stmt.ExecContext(ctx, args)
 }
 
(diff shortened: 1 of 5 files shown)


begelundmuller (changes requested):

connection.go
```diff
@@ -52,13 +65,25 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
...
 	}
 
-	if len(args) == 0 {
-		// This should be removed once duckdb_extract_statements and related APIs are available to parse multiple statements
-		// so query cancellation works for this case as well
-		return c.queryUnprepared(query)
+	stmts, size := c.extractStmts(query)
+	if size == 0 {
+		return nil, errors.New("no statements found")
 	}
```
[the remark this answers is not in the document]
pjain1: moved check to extract method

connection.go
```diff
@@ -127,33 +152,25 @@ func (c *conn) prepareStmt(cmd string) (*stmt, error) {
...
+	cquery := C.CString(query)
+	defer C.free(unsafe.Pointer(cquery))
 
-	if err == C.DuckDBError {
-		dbErr := C.GoString(C.duckdb_result_error(&res))
-		return nil, errors.New(dbErr)
-	}
+	var stmts C.duckdb_extracted_statements
+	var stmtCount C.idx_t
+	stmtCount = C.duckdb_extract_statements(*c.con, cquery, &stmts)
```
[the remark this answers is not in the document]
pjain1: added. btw extract statement does not give any error incase an empty string is give to it.

connection.go
```diff
@@ -33,17 +33,35 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
+		}
+		// send nil args to execute statement and ignore result
+		_, err = stmt.ExecContext(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		err = stmt.Close()
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 55: This will not close the statement if `ExecContext` errors

connection.go
```diff
@@ -52,17 +70,38 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
...
+	}
+	// prepare and execute last statement with args and return result
+	stmt, err := c.prepareExtractedStmt(stmts, size-1)
 	if err != nil {
 		return nil, err
 	}
 
+	C.duckdb_destroy_extracted(&stmts)
```
begelundmuller, line 104: Also needs to be defered earlier to ensure coverage of all error cases

connection.go
```diff
@@ -127,33 +166,35 @@ func (c *conn) prepareStmt(cmd string) (*stmt, error) {
...
-		dbErr := C.GoString(C.duckdb_result_error(&res))
-		return nil, errors.New(dbErr)
+	if err != "" {
+		C.duckdb_destroy_extracted(&stmts)
+		return nil, 0, errors.New(err)
+	}
+	if stmtsCount == 0 {
+		C.duckdb_destroy_extracted(&stmts)
+		return nil, 0, errors.New("no statements found")
 	}
```
begelundmuller, line 185: The docs say that `stmtsCount == 0` indicates failure. They don't say that `duckdb_extract_statements_error` is safe to call if the query didn't fail. So probably restructure this to ensure safe handling in accordance with the docs.

connection.go
pjain1: should we just ignore the error of `stmt.Close()` as even with defer we would have ignored it, it would have been printed though I think.

connection.go
pjain1: > The docs say that stmtsCount == 0 indicates failure.

I looked at the documentation, I am not sure where it explicitly says that. Even in one of the test cases [here](https://github.com/duckdb/duckdb/blob/master/test/api/capi/test_capi_extract.cpp#L48) it calls `duckdb_extract_statements_error` even if query didn't fail, so thought its ok to call in every case. Anyways I think it would be safe to do this ?
```
	if stmtsCount == 0 {
		err := C.GoString(C.duckdb_extract_statements_error(stmts))
		C.duckdb_destroy_extracted(&stmts)

		if err != "" {
			return nil, 0, errors.New(err)
		}
		return nil, 0, errors.New("no statements found")
	}
```
Let me know what you think ?

connection.go
```diff
@@ -33,13 +33,27 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
 
-	if len(args) == 0 {
-		// This should be removed once duckdb_extract_statements and related APIs are available to parse multiple statements
-		// so query cancellation works for this case as well
-		return c.execUnprepared(query)
+	stmts, size, err := c.extractStmts(query)
+	defer C.duckdb_destroy_extracted(&stmts)
+	if err != nil {
+		return nil, err
 	}
```
begelundmuller, line 40: The defer would have to come after the error check, because in those cases, `stmts` is `nil`

connection.go
```diff
@@ -52,13 +66,27 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
...
 
-	if len(args) == 0 {
-		// This should be removed once duckdb_extract_statements and related APIs are available to parse multiple statements
-		// so query cancellation works for this case as well
-		return c.queryUnprepared(query)
+	stmts, size, err := c.extractStmts(query)
+	defer C.duckdb_destroy_extracted(&stmts)
+	if err != nil {
+		return nil, err
 	}
```
begelundmuller, line 73: Same comment as above

connection.go
```diff
@@ -52,13 +66,27 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
...
+		stmt, err := c.prepareExtractedStmt(stmts, i)
+		if err != nil {
+			return nil, err
+		}
+		// send nil args to execute statement and ignore result
+		rows, err := stmt.QueryContext(ctx, nil)
+		rows.Close()
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 86: If the query failed, `rows` will be nil, so this will panic. It should probably close the `stmt` if there was an error, else close the `rows`.

(It's a little ugly that it doesn't support closing the `stmt` in all cases. Might be worth refactoring to support that in a separate PR at some point.)

begelundmuller: There are actually other cases in this code where the statement is not closed in error cases, such as the return statement in this function. It's also weird that `duckdb_destroy_prepare` is called in one error case, but not others, in `s.execute`.

Do you want to take a stab at fixing these leaks? (I think the solution is to always defer cleanup to `stmt.Close()`, and make it idempotent in case of multiple calls, which I *think* is allowed by the [spec](https://pkg.go.dev/database/sql/driver), though please double-check).

begelundmuller (changes requested):

connection.go
begelundmuller: I have attempted to address my comment above in https://github.com/marcboeker/go-duckdb/pull/91/commits/ba1759538ddc557470cb90e62c1eff17dcd433a5

begelundmuller:
@marcboeker I'm done with my reviews on this. Let us know what you think!

connection.go
```diff
@@ -33,13 +33,28 @@ func (c *conn) ExecContext(ctx context.Context, query string, args []driver.Name
...
-		// This should be removed once duckdb_extract_statements and related APIs are available to parse multiple statements
-		// so query cancellation works for this case as well
-		return c.execUnprepared(query)
+	stmts, size, err := c.extractStmts(query)
+	if err != nil {
+		return nil, err
+	}
+	defer C.duckdb_destroy_extracted(&stmts)
+
+	// execute all statements without args, except the last one
```
marcboeker, line 42: What is your idea behind preparing the last instead of the first statement with the given args? Or is this a common behaviour around `database/sql` drivers?

marcboeker (commented):
@pjain1 @begelundmuller Thanks so much for this great addition. I only have three small things otherwise we can merge.

connection.go
```diff
@@ -127,33 +159,35 @@ func (c *conn) prepareStmt(cmd string) (*stmt, error) {
...
+func (c *conn) extractStmts(query string) (C.duckdb_extracted_statements, C.idx_t, error) {
+	cquery := C.CString(query)
+	defer C.free(unsafe.Pointer(cquery))
 
-	var res C.duckdb_result
-	err := C.duckdb_query(*c.con, cmdstr, &res)
-	defer C.duckdb_destroy_result(&res)
+	var stmts C.duckdb_extracted_statements
+	var stmtsCount C.idx_t
+	stmtsCount = C.duckdb_extract_statements(*c.con, cquery, &stmts)
```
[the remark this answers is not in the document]
pjain1: its merged

connection.go
pjain1: This behaviour is same as [Python](https://github.com/duckdb/duckdb/blob/master/tools/pythonpkg/src/pyconnection.cpp#L440) and [C](https://github.com/duckdb/duckdb/blob/master/test/api/capi/test_capi_extract.cpp#L6) driver.

rows.go
```diff
@@ -262,8 +262,10 @@ func (r *rows) Close() error {
 	C.duckdb_destroy_result(&r.res)
 
 	if r.stmt != nil {
-		C.duckdb_destroy_prepare(r.stmt.stmt)
 		r.stmt.rows = false
+		if r.stmt.closeOnRowsClose {
+			r.stmt.Close()
```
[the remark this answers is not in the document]
pjain1: there is no explicit error being returned from `Close()` method on `statement`.  Except not sure if error should be actionable, may be just log the error as potential stmt leak ? @begelundmuller what are your thoughts here.

connection.go
marcboeker: Thanks for the explanation. What do you think would be the best approach?

1. Distribute all arguments over all statements so that we can do sth. like `INSERT INTO foo VALUES(?); INSERT INTO bar VALUES(?)` with two args (1, 2).
2. Only apply the args to the last statement (like it is ATM).

My preference would be 1. to be as flexible as possible and to have a predictable behaviour. Otherwise we have to document, that the given args are only applied to the last statement.

rows.go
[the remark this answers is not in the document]
begelundmuller: Fixed. Even though it can't currently return an error, I think it's a good practice to propagate errors in case the func starts returning errors in the future.

connection.go
marcboeker: Okay! I think we then should document this behaviour, to make it clear, that args are only applied to the last statement.

marcboeker:
Thanks for the contribution!

PR #95, opened 2023-05-22
Upgrade to DuckDB 0.8.0

begelundmuller:
The Linux builds initially failed on Github because the runner runs out of memory. It only happens for Linux because the Github runners for macOS have much more memory. I fixed the issue by spinning up a self-hosted runner with more memory. The builds now succeed.

After the builds succeeded, the tests failed when loading extensions on DuckDB 0.8. It turned out to be a known error and DuckDB recommended building the static libraries with `-DNDEBUG` ([details here](https://discord.com/channels/909674491309850675/921100573732909107/1110157820566962270)). After re-building with `-DNDEBUG` configured correctly all tests pass.

This PR is now ready for review.

NOTE 1: When merging, the PR should be squash merged as I had to rebuild the static libraries a few times before I got the `-DNDEBUG` config right.

NOTE 2: I have reverted the Github workflow to build DuckDB for Linux on the regular Github runners again (to avoid a dependency on my ad-hoc self-hosted runner). This means the same error may arise next time a DuckDB version is released. For a long term solution, I see two paths ahead:
- Github currently has a beta for larger runners. Use this ad-hoc self-hosted runner trick until it comes out of beta.
- Find a way to download/build the library without building the amalgamation (the explosive memory usage is a result of the C++ compiler not being optimized for large single files).

### DIFF
--- a/.github/workflows/deps.yaml
+++ b/.github/workflows/deps.yaml
@@ -38,7 +38,7 @@ jobs:
           path: deps/darwin_arm64/libduckdb.a
           retention-days: 1
   linux_amd64:
-    runs-on: self-hosted # ubuntu-20.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/deps.yaml
+++ b/.github/workflows/deps.yaml
@@ -38,7 +38,7 @@ jobs:
           path: deps/darwin_arm64/libduckdb.a
           retention-days: 1
   linux_amd64:
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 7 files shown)


marcboeker:
Awesome, thanks @begelundmuller! Great to see, that `-DNDEBUG` also reduced the file size significantly.

Regarding the amalgamation file: I'm not quite sure if DuckDB still has a focus on this or if they do not want to maintain/optimize this in the near future. Let's see how fast this [issue](https://github.com/duckdb/duckdb/issues/7539) will be fixed and if there is a future for an official and maintained amalgamation file from DuckDB.

PR #96, opened 2023-05-30
Automatic Build Support for ARM on Linux

MrFlynn:
This commit adds automatic build support for ARM64 Linux binaries. This should make it easier for people using low power ARM devices to use DuckDB with Go, but without waiting for long builds or dealing with the complexity of cross-compiling with CGo.

### DIFF
--- a/cgo_source.go
+++ b/cgo_source.go
@@ -1,4 +1,4 @@
-//go:build !duckdb_use_lib && (duckdb_from_source || !(darwin || linux))
+//go:build !(duckdb_use_lib || duckdb_from_source) && (darwin || (linux && (amd64 || arm64)))
 
 package duckdb
 
--- a/cgo_source.go
+++ b/cgo_source.go
@@ -1,4 +1,4 @@
-//go:build !duckdb_use_lib && (duckdb_from_source || !(darwin || (linux && amd64)))
+//go:build !duckdb_use_lib && (duckdb_from_source || !(darwin || linux))
 
 package duckdb
 
--- a/cgo_source.go
+++ b/cgo_source.go
@@ -1,4 +1,4 @@
-//go:build !(duckdb_use_lib || duckdb_from_source) && (darwin || (linux && (amd64 || arm64)))
+//go:build !duckdb_use_lib && (duckdb_from_source || !(darwin || (linux && (amd64 || arm64))))
 
 package duckdb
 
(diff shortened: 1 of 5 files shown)


cgo_static.go
```diff
@@ -1,4 +1,4 @@
-//go:build !(duckdb_use_lib || duckdb_from_source) && (darwin || (linux && amd64))
```
marcboeker, line 1: Same as above.

cgo_source.go
```diff
@@ -1,4 +1,4 @@
-//go:build !duckdb_use_lib && (duckdb_from_source || !(darwin || (linux && amd64)))
```
marcboeker, line 1: I think we should add `arm64` alongside `amd64` as there are plenty of other platforms on linux, we currently do not support (Mips, RISC-V)...

deps/linux_arm64/libduckdb.a
marcboeker, line 1: Please exclude all compiled libs from the commit, as they are generated by the CI and to make sure their integrity is safe.

marcboeker (commented):
Thanks for the great contribution! Please see my comments.
I'm happy to merge this after some minor modifications.

cgo_source.go
MrFlynn: I changed `(linux && amd64)` to `(linux && (amd64 || arm64))` to limit it to only those two architectures.

deps/linux_arm64/libduckdb.a
MrFlynn: I removed this binary from my branch. I also checked out the darwin binaries from master into this branch just so those weren't being modified either.

marcboeker (commented):
Thanks for the adjustments. There seems to be a small copy and paste error for the `cgo_static.go`.

cgo_source.go
```diff
@@ -1,4 +1,4 @@
...
+//go:build !(duckdb_use_lib || duckdb_from_source) && (darwin || (linux && (amd64 || arm64)))
```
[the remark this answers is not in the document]
MrFlynn: Yep, that was a copy paste error. Should be fixed now. Should also fix the failing unit tests.

cgo_static.go
```diff
@@ -1,4 +1,4 @@
...
+//go:build !duckdb_use_lib && (darwin || (linux && (amd64 || arm64)))
```
marcboeker, line 1: This condition was correct before and is now incorrect. Maybe also a copy and paste error?

PR #98, opened 2023-05-31
Better decimal support (with refactored tests)

marcboeker:


TN1ck (approved):
Thank you, looks good!

PR #101, opened 2023-06-06
Fix missing /v2 in go module path

begelundmuller:
After the v2 Git tag, I started getting errors like:

```
github.com/marcboeker/go-duckdb/v2@v2.0.1: go.mod has non-.../v2 module path "github.com/marcboeker/go-duckdb" (and .../v2/go.mod does not exist) at revision v2.0.1
```

This seems to fix it, but let me know if there's a better solution.

### DIFF
--- a/examples/simple.go
+++ b/examples/simple.go
@@ -8,7 +8,7 @@ import (
 	"log"
 	"time"
 
-	_ "github.com/marcboeker/go-duckdb"
+	_ "github.com/marcboeker/go-duckdb/v2"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


marcboeker:
Thanks for the PR. After reading [this](https://go.dev/doc/modules/major-version#considerations) I'm not quite sure anymore if switching to `v2` makes sense.  I thought that a change in the module API, which the Decimals PR definitely is, automatically results in a major version upgrade. But the impact from that would be to still maintain the `v1` and `v2`.

So maybe it makes sense, to keep `v1` and inform users about the Decimals change? What do you think?

begelundmuller:
I'm good with either approach :) I often encounter minor breaking changes in minor versions of Go packages, so wouldn't seem unusual to me.

PR #102, opened 2023-06-11
Appender: string appender errors should be returned

MathiasGr:
This change lets the error case of string appending fall through to the main error handler after the switch.

The solved issue is that with the previous version, the appender error for strings was not returned, thus impossible to troubleshoot.


MathiasGr:
Sorry about the messed up commits, let me fix that (or you can squash & merge)

marcboeker:
Thanks for cleaning up the error handling!

PR #107, opened 2023-07-28
Add a method to convert a Decimal to float64

binkynz:
Add convenience function to convert a Decimal to a float64.

### DIFF
--- a/duckdb_test.go
+++ b/duckdb_test.go
@@ -414,6 +414,14 @@ func TestDecimal(t *testing.T) {
 		require.NoError(t, db.QueryRow("SELECT 123456789.01234567890123456789::DECIMAL(29, 20)").Scan(&f))
 		compareDecimal(t, Decimal{Value: bigNumber, Width: 29, Scale: 20}, f)
 	})
+
+	t.Run("decimal to float64", func(t *testing.T) {
+		var d Decimal
+		require.NoError(t, db.QueryRow("SELECT 123.456::DECIMAL").Scan(&d))
+		f, err := d.Float64()
+		require.NoError(t, err)
+		require.Equal(t, float64(123.456), f)
+	})
 }
 
 func TestUUID(t *testing.T) {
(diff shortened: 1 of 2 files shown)


marcboeker:
Thanks for the PR. I was testing this with a larger float like `123456789.01234567890123456789` and received an error, that the precision does not match:

```
expected: 1.2345678901234567e+08
actual  : 1.23456789012e+08
```

I've also tried to circumvent the C API for conversion using

```
scale := big.NewInt(int64(d.Scale))
factor := new(big.Float).SetInt(new(big.Int).Exp(big.NewInt(10), scale, nil))
value := new(big.Float).SetInt(d.Value)
value.Quo(value, factor)
f, _ := value.Float64()
```

but got the same error as above.

After debugging I was able to see, that DuckDB incorrectly reports scale and width when it is not explicitly specified in the data type of the query. After changing the above query to `SELECT 123456789.01234567890123456789::DECIMAL(29, 20)` DuckDB correctly returns scale and width.

But using the solution via C API still results in a precision error:

```
expected: 1.2345678901234567e+08
actual  : 1.2345678901234569e+08
```

Switching back to using the Go only conversion via `big.Float` seems to work with the required precision.

Could you please [allow me to push my changes to your branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) for you to review?

binkynz:
@marcboeker "allow edit by maintainers" is selected, are you not able to push?

marcboeker:
@binkynz Ah my fault, tried to push to the wrong branch. Thanks, commits are now pushed. Please have a look at it. Thanks 🙂

binkynz:
@marcboeker lgtm and thank you

marcboeker:
@binkynz Thanks!

binkynz:
@marcboeker hey, just noticed that you are returning an error in `Float64` when an error cannot be thrown

marcboeker:
@binkynz Yeah exactly, makes no sense. Thanks for pointing this out. I've changed it.

PR #110, opened 2023-08-14
Update README.md to explain that appender creation will fail unless t…

a-agmon:
…he table mentioned will already be created

Update README.md to explain that appender creation will fail unless the table mentioned will already be created


marcboeker:
Thanks for the PR!

PR #117, opened 2023-08-22
fix: support negative decimals

TN1ck:
Just hit this problem and noticed we never tested for negative values 🫠. Each decimal we test also has a negative equivalent now.

The problem itself wasn't difficult to pinpoint as instead of e.g. -123 one got 65413, so something was off with signed/unsigned.


marcboeker:
Thanks @TN1ck 🙂 
Who needs negative numbers? /s 🙃

PR #138, opened 2023-12-21
[Fix] Memory leak with C.duckdb_struct_type_child_name

taniabogatsch:
This PR ensures that go-duckdb frees `C.duckdb_struct_type_child_name(ty, j)` after retrieving the `C.GoString`. Otherwise, this causes a memory leak of 16 bytes per function call. I also removed the `defer` to avoid warnings such as `Possible resource leak, defer is called in the for loop`.

Screenshot 2023-12-21 at 11 24 16

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -355,8 +355,13 @@ func scanList(vector C.duckdb_vector, rowIdx C.idx_t) ([]any, error) {
 
 func scanStruct(ty C.duckdb_logical_type, vector C.duckdb_vector, rowIdx C.idx_t) (map[string]any, error) {
 	data := map[string]any{}
+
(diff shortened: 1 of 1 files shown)


marcboeker:
@taniabogatsch Awesome catch, thanks for the PR!

michaelmdresser:
@taniabogatsch this is neat! How did you discover this? What tool are you using?

taniabogatsch:
@michaelmdresser, I discovered this in a scenario that runs the same query repeatedly on a table with a struct column. The memory kept growing while ignoring duckdb's memory limit until the process crashed. As for the tool, I am using Instruments, which comes as part of Xcode (which I don't use 😄). I can highly recommend it! I'm using both its Leaks analyzer, and its Time Profiler very frequently.

Issue #135, opened 2023-12-04
[Appender] Use append_data_chunk to increase performance and to support nested types


taniabogatsch:
Currently, the `Appender` operates row-at-a-time with `duckdb_append_TYPE`. However, DuckDB performs significantly better with bulk appends, i.e., with appending whole data chunks. The C API offers `duckdb_append_data_chunk`.

Additionally, there is no support for nested types in the `Appender`, and while duckdb supports appending `Value` in the C++ API (`DUCKDB_API void BaseAppender::Append(Value value)`), it does not offer the same in the C API. 

We want to pick this up and rework the `Appender` to use `duckdb_append_data_chunk` and eventually allow appending nested types by building these data chunks recursively. I've decided to open this issue to make potential other developers aware that there is work on this. I hope that's fine. :) I will close it once there is an initial PR.

Affiliation: DuckDB Labs

marcboeker:
#139 has been merged.

PR #139, opened 2023-12-22
Add Data Chunks to the Appender

maiadegraaf:
This PR alters the appender to use `duckdb::data_chunks` instead of the DuckDB [appender](https://duckdb.org/docs/api/c/appender).

The function `AppendRow()` is used the same as before, except instead of row-wise insertions, the rows are inserted into `data_chunks`, allowing us to append whole chunks at a time.  When a chunk is full, a new one is automatically allocated.  When `Flush()` is called, all the `data_chunks` are appended and destroyed in one go.

Currently [`appender_test.go`](https://github.com/marcboeker/go-duckdb/blob/master/appender_test.go) takes the same time as before (0.12s). However, I plan to tweak and optimize this function to increase its efficiency.  

Finally, a big benefit of transitioning to the `data_chunk` framework, is that it allows for the easy addition of nested types which will be coming in a future PR.

Relevant issue: #135
CC: @taniabogatsch


marcboeker:
@maiadegraaf Thanks for migrating the appender to data chunks.

killzoner:
Hey @maiadegraaf , turns out this PR probably introduces a breaking change for nested types (which were kind of already supported)

marcboeker:
@killzoner Could you please post the code you are using to insert `TEXT[]`. Thanks!

PR #144, opened 2024-01-14
Build on freebsd amd64

kytomaki:
This builds on freebsd amd64.


marcboeker:
Thanks for the PR! Do you know the compiler name for arm64 in FreeBSD so we can also offer arm64 builds?

PR #148, opened 2024-01-17
Query cancellation fix

k-anshul:
Hey @marcboeker 

Found one tricky issue in the query cancellation that we had discussed previously. 
It is easy to reproduce when query completes quickly like in case of in-memory DB and checkpoint query. Example below:

```
func main() {
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	conn, err := db.Conn(context.Background())
	if err != nil {
		log.Fatal(err)
	}

	var wg sync.WaitGroup
	wg.Add(100)
	for i := 0; i < 100; i++ {
		v := i
		go func() {
			ctx, cancel := context.WithCancel(context.Background())
			defer cancel()
			defer wg.Done()
			_, err = conn.ExecContext(ctx, "CHECKPOINT;")
			if err != nil {
				log.Printf("err %v", err)
			}
			fmt.Printf("iteration %v done\n", v)
		}()
	}
	wg.Wait()
}
```

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -203,13 +203,19 @@ func (s *stmt) execute(ctx context.Context, args []driver.NamedValue) (*C.duckdb
 	}
 	defer C.duckdb_destroy_pending(&pendingRes)
 
-	done := make(chan bool)
+	done := make(chan struct{})
 	defer close(done)
+	queryComplete := false
 
 	go func() {
 		select {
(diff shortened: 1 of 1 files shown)


marcboeker:
@k-anshul Thanks for the ping. Missed the change from draft to ready to merge.

PR #149, opened 2024-01-18
Set duckdb_api config option

elefeint:
DuckDB 0.9.2 introduced `PRAGMA USER_AGENT` consisting of the surface API (`duckdb_api`) and a composable `custom_user_agent`. 

This PR sets `duckdb_api` to `go`. And since config is now always passed when opening the database, it actually deletes more code than it adds!

### DIFF
--- a/duckdb.go
+++ b/duckdb.go
@@ -51,13 +51,11 @@ func createConnector(dataSourceName string, connInitFn func(execer driver.Execer
 		return nil, fmt.Errorf("%w: %s", errParseConfig, err.Error())
 	}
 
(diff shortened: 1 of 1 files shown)


marcboeker:
@elefeint Thanks for the PR and the refactoring. Just because I'm curious: what is the idea behind setting the `duckdb_api` identifier?

elefeint:
It allows analysis of which surfaces of DuckDB are used. Without it, `go-duckdb` usage is indistinguishable from other "C api" usage.

PR #150, opened 2024-01-19
Add Nested Types to the Appender

maiadegraaf:
I'm currently working on adding nested types to the appended.  I haven't finished it yet but I'm opening up a draft for better coordination.

#### The Big Changes:

- `STRUCT`'s and `LIST`'s are now both supported! Including nested calls:
    - `tripleNestedListInt INT[][][]`
    - `topWrapper STRUCT(Wrapper STRUCT(Base STRUCT(I INT, V VARCHAR)))`
    - `mixList STRUCT(A STRUCT(L VARCHAR[]), B STRUCT(L INT[])[])[]`
- The values are now added through recursive callback functions, initialized in `InitializeColumnTypesAndInfos`, which reduces the number of switches to one, which only happens once.

@taniabogatsch and I are still ironing out some final changes [here](https://github.com/maiadegraaf/go-duckdb/pull/1)


appender_test.go
```diff
@@ -28,7 +27,8 @@ const (
 		float REAL,
 		double DOUBLE,
 		string VARCHAR,
-		bool BOOLEAN
+		bool BOOLEAN,
+-- 	    blob BLOB
```
maiadegraaf, line 31: There previously was no test for `BLOB` types, currently it is causing a segmentation fault in 
`appender.go::443`:
```go
state = C.duckdb_append_data_chunk(*a.appender, chunk)
```

maiadegraaf: @marcboeker I've looked into this and I don't think `BLOB` types have ever fully been supported. Running this test on the current main causes a segmentation fault. To my understanding `BLOB`'s at their core look like this: `[]uint8` in golang, they overlap with slices of `uint8`, which, if I'm not missing something, means we have to choose which one to support.

I've opened an issue -> #152

appender.go
```diff
@@ -132,149 +150,329 @@ func (a *Appender) AppendRow(args ...driver.Value) error {
...
+	}
+
+	for i := 0; i < structType.NumField(); i++ {
+		childColumnInfo := columnInfo.columnInfos[i]
+		childColumnInfo.function(a, &childColumnInfo, rowIdx, refVal.Field(i).Interface())
+	}
+}
+
 // appendRowArray loads a row of values into the appender. The values are provided as an array.
 func (a *Appender) appendRowArray(args []driver.Value) error {
```
taniabogatsch, line 441: I don't think this function has to return an `error`... as of this implementation. But what happens if the user passes a value that does not match the table/schema of the first row? We need to return an actual error here, and we need to return that error if our callbacks receive a value that does not match. Can we also get a test for this?

appender.go
```diff
@@ -132,149 +150,329 @@ func (a *Appender) AppendRow(args ...driver.Value) error {
...
 		}
 		C.duckdb_destroy_data_chunk(&chunk)
 	}
+
+	// free the column types
+	for i := range a.columnInfos {
+		C.duckdb_destroy_logical_type(&a.columnTypesPtr[i])
+	}
+
+	C.free(unsafe.Pointer(a.columnTypes))
```
taniabogatsch, line 475: I think that this has to go to `Close`, no? As of now, we also call this when we call `Flush`. However, from my understanding of `Flush`, we should still be able to use the `Appender` afterward, i.e., append more chunks. Can we also have a test for this? The test appends 1-2 chunks, calls `Flush`, appends one more chunk, and then calls `Close`. Or am I misunderstanding `Flush`?

appender.go
```diff
@@ -132,149 +150,329 @@ func (a *Appender) AppendRow(args ...driver.Value) error {
...
 	return nil
 }
```
taniabogatsch, line 478: Let's throw out this unused variable.

appender.go
```diff
@@ -9,24 +9,46 @@ import (
...
-// Appender holds the DuckDB appender. It allows to load bulk data into a DuckDB database.
+type LogicalTypeEnum int
+
+const (
+	PRIMITIVE LogicalTypeEnum = iota
+	LIST
+	STRUCT
+)
+
+type SetColumnValue func(a *Appender, columnInfo *ColumnInfo, rowIdx C.idx_t, val interface{})
```
taniabogatsch, line 25: Nit: let's add a comment to this exported type.
> SetColumnValue is the type definition for all column callback functions

appender.go
```diff
@@ -9,24 +9,46 @@ import (
...
+
+type ColumnInfo struct {
```
taniabogatsch, line 27: Nit: let's add a comment to this exported type.
> ColumnInfo holds the logical column type, a callback function to write column values, and additional helper fields.

appender.go
```diff
@@ -9,24 +9,46 @@ import (
...
+	function SetColumnValue
+
+	colType LogicalTypeEnum
+
+	// Only for nested types so that we can recursively store the child column info
+	fields      int // the number of fields in the struct
+	columnInfos []ColumnInfo
+}
+
+// Appender holds the DuckDB appender. It allows data bulk loading into a DuckDB database.
```
taniabogatsch, line 38: Nit: I am tempted to change this comment to something like this:
> Appender holds the DuckDB appender. It allows efficient bulk loading into a DuckDB database.

People should use the `Appender` for bulk loading. 😄

appender.go
```diff
@@ -9,24 +9,46 @@ import (
...
 	closed   bool
 
-	currentRow       C.idx_t
 	chunks           []C.duckdb_data_chunk
-	chunkVectors     []C.duckdb_vector
-	currentChunkIdx  int
 	currentChunkSize C.idx_t
-	chunkTypes       *C.duckdb_logical_type
+	columnTypes      *C.duckdb_logical_type
+	columnTypesPtr   *[1073741823]C.duckdb_logical_type
```
taniabogatsch, line 49: I had another look at this, specifically how we allocate these C-style arrays. I think your/our approach in `initializeColumnTypes` is very similar to your approach for creating the `STRUCT` child types. My trouble with the `STRUCT` child-type approach is allocating the memory through Go, which potentially can get garbage collected. I.e., you allocate the slice in Go but then pass its data into C, pretending it to be a C-array. Is it better to allocate the memory directly in C (`C.malloc`), as we do for the column types? I found [this](https://stackoverflow.com/questions/51525876/use-go-slice-in-c). 

Additionally, I prefer something less verbose than `*[1073741823]C.duckdb_logical_type`. I had a go at this; here is my (hacky) `diff.` It also destroys the logical types for the `STRUCT` child type creation and omits the dummy logical type.

```diff
@@ -45,8 +48,8 @@ type Appender struct {
 
-       columnTypes      *C.duckdb_logical_type
-       columnTypesPtr   *[1073741823]C.duckdb_logical_type
+       columnTypes      []C.duckdb_logical_type
+       columnTypesPtr   unsafe.Pointer

@@ -265,27 +271,26 @@ func (a *Appender) InitializeColumnTypesAndInfos(v reflect.Type, colIdx int) (C.
                }
 
                structType := v
-
                columnInfo := ColumnInfo{function: f, colType: STRUCT, columnInfos: make([]ColumnInfo, v.NumField()), fields: v.NumField()}
 
-               // Create an array of the struct fields TYPES
-               // and an array of the struct fields NAMES
-               fieldTypes := make([]C.duckdb_logical_type, structType.NumField())
-               fieldNames := make([]*C.char, structType.NumField())
+               // Create an array of the struct field TYPES
+               // and an array of the struct field NAMES
+               fieldTypesPtr, fieldTypes := mallocLogicalTypeSlice(structType.NumField())
+               fieldNamesPtr, fieldNames := mallocCStringSlice(structType.NumField())
                for i := 0; i < structType.NumField(); i++ {
                        fieldTypes[i], columnInfo.columnInfos[i] = a.InitializeColumnTypesAndInfos(structType.Field(i).Type, i)
                        fieldNames[i] = C.CString(structType.Field(i).Name)
                }
 
-               fieldTypesPtr := (*C.duckdb_logical_type)(unsafe.Pointer(&fieldTypes[0]))
-               fieldNamesPtr := (**C.char)(unsafe.Pointer(&fieldNames[0]))
-
-               structLogicalType := C.duckdb_create_struct_type(fieldTypesPtr, fieldNamesPtr, C.idx_t(structType.NumField()))
+               structLogicalType := C.duckdb_create_struct_type((*C.duckdb_logical_type)(fieldTypesPtr),
+                       (**C.char)(fieldNamesPtr), C.idx_t(structType.NumField()))
 
                for i := 0; i < structType.NumField(); i++ {
+                       C.duckdb_destroy_logical_type(&fieldTypes[i])
                        C.free(unsafe.Pointer(fieldNames[i]))
-                       C.free(unsafe.Pointer(fieldTypes[i]))
                }
+               C.free(fieldTypesPtr)
+               C.free(fieldNamesPtr)

@@ -297,23 +302,18 @@
-// Create an array of duckdb types from a list of go types
+// initializeColumnTypes creates an array of DuckDB types from a list of go types
 func (a *Appender) initializeColumnTypes(args []driver.Value) {
        a.columnInfos = make([]ColumnInfo, len(args))
-       defaultLogicalType := C.duckdb_create_logical_type(0)
-       columnTypes := C.malloc(C.size_t(len(args)) * C.size_t(unsafe.Sizeof(defaultLogicalType)))
-
-       a.columnTypesPtr = (*[1<<30 - 1]C.duckdb_logical_type)(columnTypes)
+       a.columnTypesPtr, a.columnTypes = mallocLogicalTypeSlice(len(args))
 
        for i, val := range args {
                if val == nil {
                        panic(fmt.Sprintf("The first row cannot contain null values (column %d)", i))
                }
                v := reflect.ValueOf(val)
-               a.columnTypesPtr[i], a.columnInfos[i] = a.InitializeColumnTypesAndInfos(v.Type(), i)
+               a.columnTypes[i], a.columnInfos[i] = a.InitializeColumnTypesAndInfos(v.Type(), i)
        }
-
-       a.columnTypes = (*C.duckdb_logical_type)(columnTypes)
 }
 
 func (c *ColumnInfo) getChildVectors(vector C.duckdb_vector) error {

@@ -342,8 +342,8 @@  
 func (a *Appender) addChunk(colCount int) error {
        a.currentChunkSize = 0
-       // duckdb_create_data_chunk takes an array of duckdb_logical_type and a column count
-       dataChunk := C.duckdb_create_data_chunk(a.columnTypes, C.idx_t(colCount))
+       columnTypesPtr := (*C.duckdb_logical_type)(a.columnTypesPtr)
+       dataChunk := C.duckdb_create_data_chunk(columnTypesPtr, C.idx_t(colCount))
        C.duckdb_data_chunk_set_size(dataChunk, C.duckdb_vector_size())
 
        for i := 0; i < colCount; i++ {

@@ -469,11 +469,29 @@ func (a *Appender) appendChunks() error {
 
        // free the column types
        for i := range a.columnInfos {
-               C.duckdb_destroy_logical_type(&a.columnTypesPtr[i])
+               C.duckdb_destroy_logical_type(&a.columnTypes[i])
        }
 
-       C.free(unsafe.Pointer(a.columnTypes))
+       C.free(a.columnTypesPtr)
        return nil
 }
```

And here are the two new functions that I added.
```go
func mallocCStringSlice(count int) (unsafe.Pointer, []*C.char) {
	var dummyExpr *C.char
	size := C.size_t(unsafe.Sizeof(dummyExpr))

	cStringsPtr := unsafe.Pointer(C.malloc(C.size_t(count) * size))
	cStrings := (*[1 << 30]*C.char)(cStringsPtr)[:count:count]
	return cStringsPtr, cStrings
}

func mallocLogicalTypeSlice(count int) (unsafe.Pointer, []C.duckdb_logical_type) {
	var dummyType C.duckdb_logical_type
	size := C.size_t(unsafe.Sizeof(dummyType))

	columnTypesPtr := unsafe.Pointer(C.malloc(C.size_t(count) * size))
	columnTypes := (*[1 << 30]C.duckdb_logical_type)(columnTypesPtr)[:count:count]
	return columnTypesPtr, columnTypes
}
```

appender.go
```diff
@@ -54,9 +76,7 @@ func NewAppenderFromConn(driverConn driver.Conn, schema string, table string) (*
 		return nil, fmt.Errorf("can't create appender")
 	}
 
-	var chunkSize = C.duckdb_vector_size()
-
-	return &Appender{c: dbConn, schema: schema, table: table, appender: &a, currentRow: 0, currentChunkIdx: 0, currentChunkSize: chunkSize}, nil
+	return &Appender{c: dbConn, schema: schema, table: table, appender: &a, currentChunkSize: 0}, nil
```
taniabogatsch, line 79: This is unrelated to your PR, but I noticed a memory leak in this function. Also, we do not return the respective error message.

```diff
@@ -62,21 +65,26 @@ func NewAppenderFromConn(driverConn driver.Conn, schema string, table string) (*
                panic("database/sql/driver: misuse of duckdb driver: Appender after Close")
        }
 
-       var schemastr *(C.char)
+       var schemaStr *C.char
        if schema != "" {
-               schemastr = C.CString(schema)
-               defer C.free(unsafe.Pointer(schemastr))
+               schemaStr = C.CString(schema)
+               defer C.free(unsafe.Pointer(schemaStr))
        }
 
-       tablestr := C.CString(table)
-       defer C.free(unsafe.Pointer(tablestr))
+       tableStr := C.CString(table)
+       defer C.free(unsafe.Pointer(tableStr))
 
-       var a C.duckdb_appender
-       if state := C.duckdb_appender_create(*dbConn.con, schemastr, tablestr, &a); state == C.DuckDBError {
-               return nil, fmt.Errorf("can't create appender")
+       var appender C.duckdb_appender
+       state := C.duckdb_appender_create(*dbConn.con, schemaStr, tableStr, &appender)
+
+       if state == C.DuckDBError {
+               // we'll destroy the error message when destroying the appender
+               err := errors.New(C.GoString(C.duckdb_appender_error(appender)))
+               C.duckdb_appender_destroy(&appender)
+               return nil, err
        }
 
-       return &Appender{c: dbConn, schema: schema, table: table, appender: &a, currentChunkSize: 0}, nil
+       return &Appender{c: dbConn, schema: schema, table: table, appender: &appender, currentChunkSize: 0}, nil
 }
```

appender_naive_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	_, err = db.Exec(fmt.Sprintf(CREATE_TABLE, tableName))
+	checkErrWithMessage("failed to create table", err)
+
+	// insert ROW_COUNT random row
+	for i := 0; i < ROW_COUNT; i++ {
+		_, err = db.Exec(fmt.Sprintf(INSERT_INTO, tableName),
+			i, rand.Intn(10), rand.Intn(1000), rand.Intn(1000), "ducks are cool", "so are geese")
+		checkErrWithMessage("failed to insert row", err)
+	}
```
taniabogatsch, line 50: Nit: I think we should not generate random numbers in a benchmark for comparability. Let's just increment them. If you want more elaborate numbers, you can also do something like this: `(i * 95823983533) % 100000`

appender_naive_test.go
```diff
@@ -0,0 +1,83 @@
...
+        VALUES
+          (?, ?, ?, ?, ?, ?)`
+
+const ROW_COUNT = 10000
+
+func BenchmarkNaive(b *testing.B) {
+	c, err := NewConnector("", nil)
+	checkErrWithMessage("failed to connect", err)
+
+	db := sql.OpenDB(c)
```
taniabogatsch, line 38: NIt: `defer db.Close()`

appender_naive_test.go
```diff
@@ -0,0 +1,83 @@
...
+			i, rand.Intn(10), rand.Intn(1000), rand.Intn(1000), "ducks are cool", "so are geese")
+		checkErrWithMessage("failed to insert row", err)
+	}
+}
+
+func BenchmarkAppender_AppendRow(b *testing.B) {
+	c, err := NewConnector("", nil)
+	checkErrWithMessage("failed to connect", err)
+
+	db := sql.OpenDB(c)
```
taniabogatsch, line 57: Nit: `defer db.Close()`

appender_naive_test.go
taniabogatsch, line 1: Nit: should we rename this to `appender_bench_test.go`?

taniabogatsch (changes requested):
Hi @maiadegraaf, great PR! I added quite a lot of comments (again 😅), but I think that now we are almost there!

- can you double-check that we `panic` only on internal errors? In all other cases, we return an error to the user. Let's also cover all user errors with tests.
- can you double-check that we call `Close` properly on all test results (db, conn, rows, etc.) to avoid any memory leaks (I am missing destructors)? I think you already do almost everywhere, but just to be sure.

appender.go
```diff
@@ -9,24 +9,46 @@ import (
...
 )
 
-// Appender holds the DuckDB appender. It allows to load bulk data into a DuckDB database.
+type LogicalTypeEnum int
+
+const (
+	PRIMITIVE LogicalTypeEnum = iota
+	LIST
+	STRUCT
+)
```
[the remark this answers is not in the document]
maiadegraaf: I could, but I like that I know get to differentiate between primitive and nested types, although I think ultimately I only really check for nested types.

appender.go
```diff
@@ -132,149 +150,329 @@ func (a *Appender) AppendRow(args ...driver.Value) error {
...
+	refVal := reflect.ValueOf(value)
+
+	childColumnInfo := columnInfo.columnInfos[0]
+
+	childVectorSize := C.duckdb_list_vector_get_size(columnInfo.vector)
+
+	if refVal.IsNil() {
+		setNull(columnInfo, a.currentChunkSize)
+		childVectorSize += 1
+	}
```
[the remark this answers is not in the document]
maiadegraaf: Yes this works, except it is not possible to append the following in go as ints cannot be `nil`
```go
	err = appender.AppendRow(
		Int32ListList{{nil}},
	)
```

appender.go
maiadegraaf: Currently, if the user passes a value that doesn't match the table, `C.duckdb_append_data_chunk` throws an error.  The problem is that it only happens after everything has already been added into the data chunks and is only thrown once the user flushes or closes the appender. However, on the flip side, the only list of types we have here is this: `[]C.duckdb_logical_type`, which is not directly convertible to a go `Type`.  So either we would have to initialize and store an array of `Type's` to cross check, or convert the value to `C.duckdb_logical_type`.  What do you think?

appender.go
maiadegraaf: Yeah, we already do have a test for this.  In `appender_test.go`, `Flush` is called after appending 3000 rows.  But I think you're right, destroy/free should be in `Close` otherwise the logical types have to be initialized every time `Flush` is called.

appender.go
maiadegraaf: Also we only have access to the first row, which could still be different to schema, so we could throw an error for any row that doesn't match the first row, but it still could error out in `C.duckdb_append_data_chunk` if it doesn't match the schema

appender.go
```diff
@@ -132,150 +179,330 @@ func (a *Appender) AppendRow(args ...driver.Value) error {
...
+		interfaceSlice[i] = refVal.Index(i).Interface()
+	}
+
+	// Set the offset of the list vector using the current size of the child vector.
+	// The duckdb list vector contains for each row two values, the offset and the length,
+	// which are located at rowIdx*2 and rowIdx*2+1 respectively.
+	setPrimitive[uint64](a, columnInfo, rowIdx*2, uint64(childVectorSize))
+
+	// Set the length of the list vector
+	setPrimitive[uint64](a, columnInfo, rowIdx*2+1, uint64(refVal.Len()))
```
taniabogatsch, line 445: I was wondering - can we use the following struct in `abi.go` for this? And maybe move the comments there?
```Go
type duckdb_list_entry_t struct {
	offset C.idx_t
	length C.idx_t
}
```

taniabogatsch: Alternatively.
C API
```C
typedef struct {
	uint64_t offset;
	uint64_t length;
} duckdb_list_entry;
```
And something like this
```Go
		case Interval:
			val := C.duckdb_interval{
				months: C.int32_t(v.Months),
				days:   C.int32_t(v.Days),
				micros: C.int64_t(v.Micros),
			}
			if rv := C.duckdb_bind_interval(*s.stmt, C.idx_t(i+1), val); rv == C.DuckDBError {
				return errCouldNotBind
			}
```

taniabogatsch (approved):
Hi @maiadegraaf, thanks for adding all my requested changes! I only have a few comments. Let's move this PR from draft to open.

taniabogatsch:
I just saw that `go test` still fails like so:
```
tania@motorbook go-duckdb % go test
--- FAIL: TestAppenderNestedStructMismatch (0.00s)
    appender_nested_test.go:471: 
        	Error Trace:	/Users/tania/DuckDB/go-duckdb/appender_nested_test.go:471
        	Error:      	Received unexpected error:
        	            	Duckdb has returned an error while appending, all data has been invalidated.
        	            	 Check that the data being appended matches the schema.
        	            	 Error from duckdb: Invalid Input Error: Type mismatch in Append DataChunk and the types required for appender
        	Test:       	TestAppenderNestedStructMismatch
FAIL
exit status 1
FAIL	github.com/marcboeker/go-duckdb	53.019s
```

appender.go
```diff
@@ -9,24 +9,62 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 	"unsafe"
 )
 
-// Appender holds the DuckDB appender. It allows to load bulk data into a DuckDB database.
+const errNote = "NOTE: this is based on types initialized from the first row of appended data, please confirm this matches the schema."
```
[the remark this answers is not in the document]
maiadegraaf: #155

maiadegraaf:
Thanks for your feedback! I've implemented all the requested changes, so the PR is ready to be merged!

marcboeker:
@maiadegraaf Thanks for the PR. Just a quick update: I'm currently refactoring this a bit.

marcboeker:
First iteration of refactorings; more will follow.

taniabogatsch:
@marcboeker, what motivates [removing significant test coverage](https://github.com/marcboeker/go-duckdb/pull/150/commits/071952af766ebab57622d0f18480cbc3a6d9f645) to have the tests run faster? Especially since these cover the new nested functionality?

marcboeker:
> @marcboeker, what motivates [removing significant test coverage](https://github.com/marcboeker/go-duckdb/pull/150/commits/071952af766ebab57622d0f18480cbc3a6d9f645) to have the tests run faster? Especially since these cover the new nested functionality?

Please review the `appender_test.go` file, where nested and base cases have been consolidated and refactored (though not yet complete). The only test removed is the nested large test, as I believe the base nested test is sufficient for testing with e.g. 10 rows instead of 10000, without sacrificing the speed of the test suite. Rapid tests are essential for working with them.

Although benchmarks can be helpful, they are not present for the rest of go-duckdb. Therefore, I have removed them. My aim is to create test cases that cover a significant portion of the code, are comprehensible, and can serve as a form of documentation.

There are still missing test cases for the appender, such as UUID.

taniabogatsch:
Thanks for the explanation. 

> The only test removed is the nested large test, as I believe the base nested test is sufficient for testing with e.g. 10 rows instead of 10000

The reason for having more than ten rows is that DuckDB's standard vector size is 2048. I see that the primitive types now cover 3k rows, which tests the creation of multiple chunks. I missed that change. Child vectors can exceed the standard vector size for nested types, so testing this might still be interesting.

> without sacrificing the speed of the test suite. Rapid tests are essential for working with them.

It is possible to use a regex to run only specific tests, and it is also possible to group slow tests into a common expression to exclude (e.g., `_slow_test.go`). That way, it is possible to include tests closer to real-usage scenarios while keeping fast tests for rapid development.

This is mostly me trying to understand the project better and better understand the structure for future contributions. Maybe it is worth setting up a section in the [contributing guidelines](https://github.com/marcboeker/go-duckdb/blob/main/CONTRIBUTING.md) about this?

taniabogatsch:
I had a look at the failing test. The reason for the segmentation fault is that we do not set the validity mask correctly for the child vectors of the `STRUCT` vector. When appending the data, we assume the values are set and try to append a string with an undefined length and an invalid data pointer.

I guess this worked on macos but not on Ubuntu because Ubuntu probably default-initialised that bit to one (valid). In contrast, MacOS probably default-initialised that bit to zero (false)... Or some other bit-level shenanigans.

I have a fix ready, and I'm also working on some other changes. I'll open a PR to @maiadegraaf once I'm done.

I'll also get back to you about the `BLOB`s.

marcboeker:
@taniabogatsch Thanks, sounds great. I'm curious to see, how you have fixed it.

taniabogatsch:
I used the logical type to switch in the `SetNull,` but I believe that only works for top-level recursion depths. We either have to destroy all child logical types when initializing the types or when closing the `Appender`. So these types might not be accessible in deeper levels. A cleaner solution here is to revert to the original implementation of this PR, where we return the logical type and end up with exactly one logical type for the top-level chunk initialization. That way, we also don't keep an unused field in the `colInfo` struct, and we can instead set the `DUCKDB_TYPE`. I'll push a fix. 🤔

taniabogatsch:
W.r.t. the `BLOB` type. DuckDB distinguishes between `BLOB` and `UTINYINT[]`. In DuckDB, a `BLOB` has the same representation as a `VARCHAR`. Both are a more elaborate char pointer (inlined vs. string heap). It is possible to store `Go-string` and `Go-[]byte` in `C.CString`, and pass the resulting C string to `duckdb_vector_assign_string_element`. I updated the `BLOB` implementation accordingly, and I also expanded on the test.

To the best of my knowledge, Go does not distinguish between `[]byte` and `[]uint8`. DuckDB distinguishes between `BLOB` and `UTINYINT[]`. @maiadegraaf raised this problematic [here](https://github.com/marcboeker/go-duckdb/pull/150#discussion_r1458674212). In the current solution, we opt to support `BLOB` in the `Appender`, but not `UTINYINT[]`. I added the respective tests.

cc @marcboeker, what do you think? :)

appender.go
```diff
@@ -9,24 +9,85 @@ import (
...
+// SetColValue is the type definition for all column callback functions.
+type SetColValue func(a *Appender, info *colInfo, rowIdx C.idx_t, val any)
+
+// colInfo holds a column's underlying vector, a callback function to write column values to this vector, and additional helper fields.
+type colInfo struct {
+	vector C.duckdb_vector
+	fn     SetColValue
+
+	// The type of the column.
+	ddbType C.duckdb_type
```
marcboeker, line 48: I would call it `duckdbType` or `dbType` as it is more clear than `ddbType`. Reads like a typo.

appender.go
```diff
@@ -40,23 +101,32 @@ func NewAppenderFromConn(driverConn driver.Conn, schema, table string) (*Appende
...
-	defer C.free(unsafe.Pointer(tablestr))
+	cTable := C.CString(table)
+	defer C.free(unsafe.Pointer(cTable))
 
-	var a C.duckdb_appender
-	if state := C.duckdb_appender_create(*dbConn.con, schemastr, tablestr, &a); state == C.DuckDBError {
-		return nil, fmt.Errorf("can't create appender")
-	}
+	var appender C.duckdb_appender
+	state := C.duckdb_appender_create(*dbConn.con, cSchema, cTable, &appender)
```
marcboeker, line 114: Maybe we should shorten `dbConn` simply to `conn` as the corresponding struct field in the `Appender` is also called `c`.

appender.go
```diff
@@ -66,214 +136,465 @@ func (a *Appender) Error() error {
...
+func mallocCStringSlice(count int) (unsafe.Pointer, []*C.char) {
+	var dummy *C.char
+	size := C.size_t(unsafe.Sizeof(dummy))
+
+	csPtr := unsafe.Pointer(C.malloc(C.size_t(count) * size))
+	slice := (*[1 << 30]*C.char)(csPtr)[:count:count]
+	return csPtr, slice
+}
+
+func initPrimitive[T any](ddbType C.duckdb_type) (colInfo, C.duckdb_logical_type) {
```
marcboeker, line 247: Awesome use of generics. Makes the code a lot cleaner 🎉
Maybe we should rename `ddbType` to `duckdbType` here to.

marcboeker (commented):
Thanks for the improvements. I really like the generics approach. I'm going to merge this and we can fix the small changes later.

PR #153, opened 2024-01-24
destroy config when closing connector

levakin:
This PR delegates destroy config call to Close method of the Connector.
This changes were extracted from the #151 
cc @taniabogatsch @marcboeker


marcboeker:
@levakin Thanks for the PR. Closing this in favour of #154.

PR #156, opened 2024-02-01
add sql null types

scottlepp:


marcboeker:
Thanks for the PR @scottlepp. In PR #150 a complete refactoring of the Appender is introduced, which will probably beak your PR. I suggest that you check your modifications again, when #150 has been merged (probably this week).

scottlepp:
Hi @marcboeker.  I'm using that fork now.  That allows appending nil values.  Though the first value in the vector can't be null.  I'm assuming that is because it needs a value to determine the datatype for the vector.  I have a workaround but does that sound right?

maiadegraaf:
@scottlepp Yes exactly, the first row is used to initialize the types, and because if the user tries to append the first row like this:
```go

err = appender.AppendRow(
		nil,
	)

```
there is no way to derive the type so it can't be properly initialized.

PR #160, opened 2024-02-14
upgrade to 0.10.0

esevastyanov:


marcboeker:
Thanks @esevastyanov for the PR. There is currently another PR #163 open, that bumps the version to 0.10.0 and adds the stdlib.h as an additional include, as DuckDB removed it from their build. We're currently trying to fix another issue with 0.10.0 and will be merging the other PR shortly. So I'm closing this.

Issue #155, opened 2024-01-31
After next go-duckdb release: Use New C API function to get Types from the Appender duckdb/duckdb#10401


maiadegraaf:
The next DuckDB release will include a function that returns the column types from the appender (https://github.com/duckdb/duckdb/pull/10401). The appender needs to be altered to check if the types being added to the data chunk match those of the table.

taniabogatsch:
PR: #171

taniabogatsch:
This issue is done.

PR #168, opened 2024-02-19
feat(appender): Allow typed nils in first row, dereference pointers

disq:
Removes the requirement of having non-null columns in the first row, by using typed nils (and dereferencing pointers if necessary)

The check for typed nils (using reflection) may be considered an expensive one, I would prefer if `initColTypes` was exported, then it can be called separately (optionally) with concrete types if caller knows the first row _may_ contain nulls, then we can remove the typed nils and the check for it (and close this PR without merging)

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -257,6 +257,10 @@ func initPrimitive[T any](ddbType C.duckdb_type) (colInfo, C.duckdb_logical_type
 }
 
 func (a *Appender) initColInfos(v reflect.Type, colIdx int) (colInfo, C.duckdb_logical_type) {
+	if v.Kind() == reflect.Ptr {
+		return a.initColInfos(v.Elem(), colIdx)
+	}
+
 	switch v.Kind() {
 	case reflect.Uint8:
 		return initPrimitive[uint8](C.DUCKDB_TYPE_UTINYINT)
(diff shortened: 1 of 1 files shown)


disq:
cc @taniabogatsch please can you also take a look?

taniabogatsch:
Hi @disq! Thanks for the PR - the appender/data chunk types must match the types in the duckdb table. As of duckdb's C API version 0.10., we can remove inferring the types from the first row altogether. @maiadegraaf opened a respective issue [here](https://github.com/marcboeker/go-duckdb/issues/155). This might prove cleaner than the current solution, and I'd also prefer it to your changes. What do you think? 

```c
DUCKDB_API idx_t duckdb_appender_column_count(duckdb_appender appender);
DUCKDB_API duckdb_logical_type duckdb_appender_column_type(duckdb_appender appender, idx_t col_idx);
```

After creating the appender, you must first retrieve the column count. Then, for each column, you retrieve the type. Then, you have all the necessary types to initialize the data chunks.

Let me know if you pick this up. Otherwise, I'll eventually have a go at it.

disq:
@taniabogatsch @maiadegraaf Thanks for your input. Yeah, requiring typed nils also require some kind of set up in the caller (as typed nils are considered uncommon) I'd prefer the new approach as well. I don't think I can pick this up this week, will let you know if I do.

Closing PR, but I'll hang onto the branch in my local for now.

taniabogatsch:
PR is up.

PR #167, opened 2024-02-19
Remaining Appender review changes and test fixes

taniabogatsch:
This PR contains the changes mentioned in the remaining comments of #150. Additionally, I noticed that the tests missed `defer Close()`-statements. There should be no more leaks in the test cases now.

Issue #89, opened 2023-04-28
Append rows using sql.Null* types


mihelich:
`Appender.AppendRowArray` doesn't recognize the `sql.Null*` value types like [sql.NullString](https://pkg.go.dev/database/sql#NullString). Clients can provide a `driver.Value` that is either `nil` (to insert NULL) or a fundamental type, but the `sql.Null*` types are more readable and type-safe.

mlafeldt:
👋 Thanks for opening this issue!

This issue has been moved to https://github.com/duckdb/duckdb-go/issues/11, and I'm closing it here.

Please see https://github.com/marcboeker/go-duckdb/issues/565 for more details.

PR #171, opened 2024-02-19
Initialise column types during appender creation

taniabogatsch:
As discussed in #168, this PR adds support for initializing the column types of a data chunk when creating an appender.

The PR builds on https://github.com/marcboeker/go-duckdb/pull/167. I'll undraft it once #167 is merged.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -46,7 +46,7 @@ type colInfo struct {
 	fn     SetColValue
 
 	// The type of the column.
-	ddbType C.duckdb_type
+	duckdbType C.duckdb_type
 	// The number of fields in a STRUCT column.
 	numFields int
 	// Recursively stores the child colInfos for nested types.
@@ -54,12 +54,12 @@ type colInfo struct {
 }
 
 func (c *colInfo) duckDBTypeToString() string {
-	if c.ddbType == C.DUCKDB_TYPE_LIST {
+	if c.duckdbType == C.DUCKDB_TYPE_LIST {
 		s := c.colInfos[0].duckDBTypeToString()
 		return "[]" + s
 	}
 
-	if c.ddbType == C.DUCKDB_TYPE_STRUCT {
+	if c.duckdbType == C.DUCKDB_TYPE_STRUCT {
 		s := "{"
 		for i := 0; i < c.numFields; i++ {
 			if i > 0 {
@@ -72,7 +72,7 @@ func (c *colInfo) duckDBTypeToString() string {
 		return s
 	}
 
-	return typeIdMap[c.ddbType]
+	return typeIdMap[c.duckdbType]
 }
 
 // Appender holds the DuckDB appender. It allows efficient bulk loading into a DuckDB database.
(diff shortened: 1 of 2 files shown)


taniabogatsch:
I added a map to track all as-of-now unsupported types in the appender. We can incrementally add these (if they make sense in Go).

appender.go
```diff
@@ -562,7 +548,7 @@ func (a *Appender) appendRowArray(args []driver.Value) error {
 		}
 
 		if err := info.typeMatch(reflect.TypeOf(v)); err != nil {
-			return fmt.Errorf("type mismatch for column %d: \n%s \n%s", i, err.Error(), errNote)
+			return fmt.Errorf("type mismatch for column %d: \n%s", i, err.Error())
```
disq, line 551:
```suggestion
			return fmt.Errorf("type mismatch for column %d: \n%s", i + 1, err.Error())
```

I think 1-indexed would be more clear since we're talking about columns and not slice elements in the error message.

disq:
@taniabogatsch in `appendRowArray`, the untyped nil check (quoted below) doesn't seem to be enough for nullable list items.

```go
		if v == nil {
			setNull(&info, a.currSize)
			continue
		}
```

To be able to append a list with a null value in it, the slice has to be of type `[]*T`. But `typeMatch` won't allow it:

> type mismatch for column 29: expected: []int64, actual: []*int64

```
select column_name, ordinal_position, is_nullable, data_type from information_schema.columns where ordinal_position=30;
┌─────────────┬──────────────────┬─────────────┬───────────┐
│ column_name │ ordinal_position │ is_nullable │ data_type │
│   varchar   │      int32       │   varchar   │  varchar  │
├─────────────┼──────────────────┼─────────────┼───────────┤
│ id_list     │               30 │ YES         │ BIGINT[]  │
└─────────────┴──────────────────┴─────────────┴───────────┘
D 
```

appender.go
taniabogatsch: Fair point! I've updated both error messages (append row column type mismatch and table column type mismatch during appender creation).

appender.go
```diff
@@ -236,169 +230,188 @@ func mallocLogicalTypeSlice(count int) (unsafe.Pointer, []C.duckdb_logical_type)
...
+}
 
-	csPtr := unsafe.Pointer(C.malloc(C.size_t(count) * size))
-	slice := (*[1 << 30]*C.char)(csPtr)[:count:count]
-	return csPtr, slice
+func getValue[T any](val any) T {
+	if reflect.ValueOf(val).Kind() != reflect.Ptr {
+		return val.(T)
+	}
+	return *(val.(*T))
```
disq, line 248: There's a nil pointer dereference here, called from the `colInfo` generated in `initCString` (`blob` branch) and called from `setList`'s `childInfo.fn`

taniabogatsch: Nice catch - I was missing an early out. On this note, we need to add more testing to these changes. We now allow passing pointers also for top-level types. I am already extending some test cases, but please drop anything here that you can find that still breaks!

taniabogatsch:
Also, cc @marcboeker, what is your take on allowing (nil)-pointers as types in the appender? I understand it is an elegant solution to better support `NULL` types for now. Is there another way? We could potentially also move to the [`database/sql`](https://pkg.go.dev/database/sql#NullInt32) package `NULL` types. 🤔 

That is, we focus on column type initialization in this PR. Then, we can approach the whole `NULL` topic in a separate, dedicated PR. As I already noticed other tests breaking, if I extend them to this solution. cc @disq ?

appender.go
disq: Our test suite (of various types and lists of those types) is now passing without issue, thanks for the rapid responses.

marcboeker:
#167 is merged now 🙂

taniabogatsch:
@disq - fair point! I'll disentangle this PR to focus on the type creation/initialization. Then, I'll open a separate PR and port my `NULL` changes to that one. That way, we do not mix both, and I can focus on adding a few more tests for the `NULL`s. Eventually, we can discuss better `sql.Null*` type support in the appender/go-duckdb, as raised in https://github.com/marcboeker/go-duckdb/issues/89.

PR #180, opened 2024-03-05
TIMESTAMP_TZ support

### LINKED ISSUE #172: Tables containing columns of type TIMESTAMPTZ cannot be queried.
Hello,
after updating to go-duckdb v1.6.1, if the table contains columns of type `TIMESTAMPTZ`, then the table cannot be queried properly and an error is returned: `unsupported type 32`. 
Can you confirm if this is a bug or if there are other types of columns that can be used in place of `TIMESTAMPTZ`. 
go-duckdb v1.5.6 is able to query `TIMESTAMPTZ` type columns properly.

taniabogatsch:
This PR adds support for `TIMESTAMP_TZ`. I assume that we treated them equally in the C API in a previous go-duckdb version. Therefore, the latest release broke compatibility.

Fixes #172.

@hawkfish, could you maybe review my attempt at a `README.md` section to outline the behavior?

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ defer db.Close()
 
 Please refer to the [database/sql](https://godoc.org/database/sql) documentation for further usage instructions.
 
-## A Note on Memory Allocation
+## Memory Allocation
 
 DuckDB lives in-process. Therefore, all its memory lives in the driver. All allocations live in the host process, which
 is the Go application. Especially for long-running applications, it is crucial to call the corresponding `Close`-functions as specified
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -178,3 +178,12 @@ LD_LIBRARY_PATH=/path/to/libs ./main
...
 DYLD_LIBRARY_PATH=/path/to/libs ./main
 ```
+
+## Notes
+
+`TIMESTAMP vs. TIMESTAMP_TZ`
+
+In the C API, DuckDB stores both `TIMESTAMP` and `TIMESTAMP_TZ` as `duckdb_timestamp`, which holds the number of 
+microseconds elapsed since January 1, 1970 UTC. When passing a `time.Time` with a different location to go-duckdb, 
+it'll transform it with `UnixMicro()`, even when using `TIMESTAMP_TZ`. Scanning that value later returns the time in `UTC`,
```
hawkfish, line 188: In the C API, DuckDB stores both `TIMESTAMP` and `TIMESTAMP_TZ` as `duckdb_timestamp`, which holds the number of 
microseconds elapsed since January 1, 1970 UTC (i.e., an instant without offset information). When passing a `time.Time` to go-duckdb, 
it will transform it to an instant with `UnixMicro()`, even when using `TIMESTAMP_TZ`. Later, scanning either type of value returns an instant, as SQL types do not model time zone information for individual values.

hawkfish (commented):
Changed the TZ description to use instants instead of UTC.

niger-prequel:
I had [opened my own based](https://github.com/marcboeker/go-duckdb/pull/181#issue-2170082841) on what I was doing locally but if you don't see anything in mine not covered here just close it.

marcboeker:
@taniabogatsch Thanks for the PR. LGTM.

PR #181, opened 2024-03-05
Fix TIMESTAMP_TZ issue in 0.10.0

### LINKED ISSUE #172: Tables containing columns of type TIMESTAMPTZ cannot be queried.

niger-prequel:
Fixes #172 

### Completed

- [x] Update duckdb.h
- [x] Include serialization for `DUCKDB_TYPE_TIMESTAMP_TZ`

### Notes

- There may be other types that need similar treatment like `DUCKDB_TYPE_TIME_TZ`. I only addressed what was blocking functionality for Prequel
- I believe this only works against the HEAD of `main` for DuckDB and not the `v0.10.0` tag.

### DIFF
--- a/duckdb.h
+++ b/duckdb.h
@@ -64,74 +64,77 @@ extern "C" {
 //===--------------------------------------------------------------------===//
 // Enums
 //===--------------------------------------------------------------------===//
-
+// WARNING: the numbers of these enums should not be changed, as changing the numbers breaks ABI compatibility
+// Always add enums at the END of the enum
 //! An enum over DuckDB's internal types.
 typedef enum DUCKDB_TYPE {
 	DUCKDB_TYPE_INVALID = 0,
 	// bool
-	DUCKDB_TYPE_BOOLEAN,
+	DUCKDB_TYPE_BOOLEAN = 1,
 	// int8_t
-	DUCKDB_TYPE_TINYINT,
+	DUCKDB_TYPE_TINYINT = 2,
 	// int16_t
-	DUCKDB_TYPE_SMALLINT,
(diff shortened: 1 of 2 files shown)


niger-prequel:
@taniabogatsch reverted the header change.

niger-prequel:
Actually @taniabogatsch I do think we need the header changes

```
#0 26.51 /usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
#0 26.51 /usr/bin/ld: /tmp/go-link-2075944908/000038.o: in function `_cgo_c4075e040998_Cfunc_duckdb_appender_column_count':
#0 26.51 appender.cgo2.c:(.text+0xb2): undefined reference to `duckdb_appender_column_count'
#0 26.51 /usr/bin/ld: /tmp/go-link-2075944908/000038.o: in function `_cgo_c4075e040998_Cfunc_duckdb_appender_column_type':
#0 26.51 appender.cgo2.c:(.text+0x109): undefined reference to `duckdb_appender_column_type'
#0 26.51 collect2: error: ld returned 1 exit status
```

I get the above trying to run this with the head of DuckDB

marcboeker:
@niger-prequel Thanks for the PR. I think we can close this in favor of #180?
cc @taniabogatsch

PR #183, opened 2024-03-06
Scan type and a type name of TIMESTAMPTZ

esevastyanov:
Complements #180


marcboeker:
@esevastyanov Thank you!

PR #187, opened 2024-03-20
Bump Go version, unify errors and other refactoring

taniabogatsch:
### New error functions
This PR adds a new file, `errors.go`, which unifies driver error messages by introducing two new functions: `getError` and `getDuckDBError`. For now, I've only used the new error functions in `duckdb.go`, but if this is a sensible proposal, I'll open more PRs to unify errors throughout the driver.

### Other changes
@marcboeker, let me know what you think. :)
- `con    *C.duckdb_connection` becomes `duckdbConn C.duckdb_connection`, as it already is a pointer.
- renamed the type `conn` to `connection`, to avoid shadowing when using the variable name `conn`.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -107,7 +107,7 @@ func NewAppenderFromConn(driverConn driver.Conn, schema, table string) (*Appende
 	defer C.free(unsafe.Pointer(cTable))
 
 	var appender C.duckdb_appender
-	state := C.duckdb_appender_create(c.duckdbConn, cSchema, cTable, &appender)
+	state := C.duckdb_appender_create(c.duckdbCon, cSchema, cTable, &appender)
 
 	if state == C.DuckDBError {
 		// We'll destroy the error message when destroying the appender.
--- a/appender.go
+++ b/appender.go
@@ -72,7 +72,7 @@ type colInfo struct {
 
 // Appender holds the DuckDB appender. It allows efficient bulk loading into a DuckDB database.
 type Appender struct {
-	c        *conn
+	c        *connection
 	schema   string
 	table    string
 	appender *C.duckdb_appender
@@ -88,7 +88,7 @@ type Appender struct {
 
 // NewAppenderFromConn returns a new Appender from a DuckDB driver connection.
 func NewAppenderFromConn(driverConn driver.Conn, schema, table string) (*Appender, error) {
-	c, ok := driverConn.(*conn)
+	c, ok := driverConn.(*connection)
 	if !ok {
 		return nil, fmt.Errorf("not a duckdb driver connection")
 	}
@@ -107,7 +107,7 @@ func NewAppenderFromConn(driverConn driver.Conn, schema, table string) (*Appende
 	defer C.free(unsafe.Pointer(cTable))
 
 	var appender C.duckdb_appender
-	state := C.duckdb_appender_create(*c.con, cSchema, cTable, &appender)
+	state := C.duckdb_appender_create(c.duckdbConn, cSchema, cTable, &appender)
 
 	if state == C.DuckDBError {
 		// We'll destroy the error message when destroying the appender.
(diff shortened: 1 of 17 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/marcboeker/go-duckdb
 
-go 1.18
+go 1.21
```
marcboeker, line 3: If we switch the Go versions, is there a reason not to go with 1.22.1?

appender.go
```diff
@@ -72,7 +72,7 @@ type colInfo struct {
 
 // Appender holds the DuckDB appender. It allows efficient bulk loading into a DuckDB database.
 type Appender struct {
-	c        *conn
+	c        *connection
```
marcboeker, line 75: I'm not married to `conn` but I've thought we could use the standard abbreviation in Go for a connection.

https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/iprawsock.go;l=99
https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/tcpsock.go;l=112
https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.21.0:ssh/client.go;l=20

connection.go
```diff
@@ -15,21 +15,21 @@ import (
 	"unsafe"
 )
 
-type conn struct {
-	con    *C.duckdb_connection
-	closed bool
-	tx     bool
+type connection struct {
+	duckdbConn C.duckdb_connection
```
marcboeker, line 19: That's a good improvement and makes it clearer than is currently is.

errors_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+var testErrOpenMap = map[string]string{
+	errParseDSN.Error():  ":mem ory:",
+	errOpen.Error():      "?readonly",
+	errSetConfig.Error(): "?threads=NaN",
+}
+
+func TestErrOpen(t *testing.T) {
+
```
marcboeker, line 16: I would suggest to go with gofumpt's rule no 2:

https://github.com/mvdan/gofumpt?tab=readme-ov-file#added-rules

I always use gofumpt instead of gofmt, as it is stricter and it alwys removes blank lines at the beginning of a function body.

marcboeker (commented):
Thanks for the refactoring and centralized errors. I've added some minor comments.

errors_test.go
taniabogatsch: Let's make it a formatting rule, then. The new line after the method declaration is a personal preference, i.e., it increases readability for me, but in the end I don't mind. I'll try to add a check/CI formatter for it.

taniabogatsch: Okay, I've looked a bit into `gofumpt` but I only ran it on this branch. I'll add an issue, maybe someone wants to pick it up. I couldn't find a pluggable CI action like for `golangci-lint`, so we'd have to add it to the CI ourselves.

appender.go
taniabogatsch: Fair point. We also have it in `driver.Conn`. I was mainly trying to unify this, as I was getting confused by our different naming conventions: `connection, con, conn, c, Connector...`. 

Maybe we can unify it like this?
```Go
type conn
var con conn

type Connector
var c Connector
```

taniabogatsch:
It looks like our `freebsd` tests cannot be compiled to `1.22`, so I'll revert back to 1.21.
```
  gmake test
  /usr/bin/bash /home/runner/work/_actions/vmactions/freebsd-vm/v1/run.sh execSSHSH
  Config file: freebsd-14.0.conf
  go test -v -race -count=1 .
  go: downloading github.com/apache/arrow/go/v14 v14.0.2
...
  go: downloading github.com/klauspost/cpuid/v2 v2.2.5
  github.com/marcboeker/go-duckdb: cannot compile Go 1.22 code
  FAIL	github.com/marcboeker/go-duckdb [build failed]
  FAIL
  gmake: *** [Makefile:14: test] Error 1
```

appender.go
taniabogatsch: I've reverted the changes, but I'd love it if we could slowly move towards more uniform names for this situation in future PRs. :)

marcboeker:
Thanks for the modifications.

PR #201, opened 2024-04-27
Table function UDFs

JAicewizard:
This implements table function UDFs.

The function takes in an interface, from which it detects the type of the parameters and return values.
This interface also has a method used to generate new rows

The interface is quite simple, but not ideal. There is an issue that we need to obtain the type of the UDF from the user-input, but we also don't want the user to specify the duckdb types manually. This solution allows the user to return arbitrary types, and they will be matched to duckdb types automatically.

Another limitation is that the value interface only allows retrieving string and int64 types. This is only a limitation for the arguments, however it is a large limitation.


JAicewizard:
PS, dont merge yet, I would like a review but c allocations are handled sloppily, and often not de-allocated. Also more return types can (and should) be implemented still

marcboeker:
@JAicewizard Thanks for the contribution. I think it makes sense to also add a lot of test cases for this feature.

ajzo90:
Great initiative! Based on this contribution I have experimented with a different API for my own needs. In particular it support predicate pushdown, concurrent scanners, and exposes an API to work with table states. 

[udf.go](https://github.com/ajzo90/go-duckdb/blob/master/udf.go)
[example](https://github.com/ajzo90/go-duckdb/blob/master/examples/udf/udf.go)

JAicewizard:
@ajzo90 Those are some great changes, one thing I didn't want to do however was add a bunch of `Chunk` and `Vector` APIs to this library. I agree that they would be great, I think leaving this PR to as un-vectorised, we cal always add a vectorised version later.

In the design of the `Chunk` and `Vector`apis it would be wise to also think of a way we might be able to vectorise the `Appender` API at the same time. (The scanner type for example looks a lot like something that could be used in the appender).
The first thing I will do it create benchmarks to see what the bottle-neck is, my suspicion is that it is the conversion to `any` for every value, which should be removable.

I will however definitely look at how you did predicatepushdown, I do not really know what this actually does so I didn't touch it.

JAicewizard:
> @JAicewizard Thanks for the contribution. I think it makes sense to also add a lot of test cases for this feature.

I will, and a benchmark as well. This APi is very much not optimal, but I am glad you're open to this PR :)

JAicewizard:
I implemented a generic interface for the `vector` type which allows me to speed up the UDF by 3x, as it no longer needs to push the values into an interface (at least for primitive types).

However it adds some new functions in `appender_vector.go`, so I would like a quick look at this code. (note that go does not allow generic parameters on methods).

If all is well I can implement some of the features from ajzos branch to make this feature complete!

JAicewizard:
todo list:

- [ ] named parameters
- [ ] max threads
- [ ] testing of projection pushdown
- [ ] making sure all allocations are freed

JAicewizard:
I force-pushed to cleanup the history. This PR ended up growing a bit out of hand, especially due to the wanting a safe API for setting the types of columns, without forcing the user to pass us a value. I think the current version is nice, as it the duckdb logical type is fully managed by go-duckdb, while also giving users a nice API.

One thing I am not entirely happy with is the integration with the existing `vector` type, however I did not feel the need to implement a fully custom one just for table functions. The added code allows for setting values without having to turn them into an `any`. This _significantly_ speeds up table functions, as converting a value into `any` does a heap allocation which is expensive.

I think this ready for a review.

taniabogatsch:
Hi @JAicewizard, with the release of duckdb 0.10.3, we now also expose scalar UDFs in the C API: https://github.com/duckdb/duckdb/pull/11786. I am currently looking into adding support for this to go-duckdb. Your PR is a great help! 😊 I noticed that there are many functions that both implementations could benefit from, like your changes to `type.go`. I am wondering on how to best organise efforts to avoid writing similar code twice. 🤔 Maybe I could work off of your PR, or maybe you have other suggestions?

taniabogatsch:
I am also open to finalising this PR first, and then starting on the scalar UDFs. In that case, I can give a thorough review, or open a PR to your PR with suggestions? Let me know what you think!

taniabogatsch:
Ignore part of my comments, and sorry for any confusion. The scalar UDFs are on our `feature` branch, so not yet part of the release. However, they will be in the not-too-far future, haha. So I'll just draft the scalar UDFs, and review this PR with that in mind. 😄

JAicewizard:
haha I was confused about scalar UDFs in 10.3 already. I indeed wrote much of the code in a way that in the future scalar UDFs would also be able to benefit from the infrastructure. Thats also why I put the `Type` and `Row` in separate files.

I was also thinking about doing scalar UDFs, but you can take that on if you want. I think much of the code can be copied/used for inspiration, but I'm afraid it will be difficult to make much of the callback code shared. Most code is handling parameters and returned values which can't be shared.

After this PR I will start working on another one adding vectorised table UDFs, do you have any ideas on how to implement a nice and fast way to represent a duckdb vector? I was thinking maybe something with a method `SetValues[T Any](data []T)` which sets the first `len(data)` values to the specified ones, and clears the others. But I am open for ideas!

A review would definitely be highly appreciated, I can review your code as well once its done!

type.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			var err error
+			(*types)[i], err = tryGetDuckdbTypeFromValue(rt.Field(i).Type)
+			if err != nil {
+				return C.duckdb_create_logical_type(C.DUCKDB_TYPE_INVALID), err
+			}
+			(*names)[i] = C.CString(rt.Field(i).Name)
+		}
+		ctypes := (*C.duckdb_logical_type)(unsafe.Pointer(types))
+		cnames := (**C.char)(unsafe.Pointer(names))
```
JAicewizard, line 173: PS, I think there is a bug here, I will write a test and fix it soon

JAicewizard:
@taniabogatsch I renamed the files to `udtf*.go` so that you can create new ones.
It might also be nice for scalar UDFs to be able to extract the type information from the function being passed like in python.

ajzo90 (commented):
Nice job with the UDF implementation.
I left a few comments related to projection pushdown and threading. In my opinion it make sense to wait with these (performance related) features until we know where we want to go with vectorised APIs.

udf.go
```diff
@@ -0,0 +1,259 @@
...
+		MaxThreads int
+	}
+
+	tableFunctionData struct {
+		fun        TableFunction
+		projection []int
+	}
+
+	TableFunction interface {
+		Init() TableFunctionInitData
```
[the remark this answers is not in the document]
JAicewizard: Good point, that would be inline with what happens when executing table functions, the functions get the info, and can do with it what they want (ignore it or skip the unused columns)

udf.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	CardinalityData struct {
+		Cardinality uint
+		IsExact     bool
+	}
+
+	TableFunctionInitData struct {
+		MaxThreads int
+	}
```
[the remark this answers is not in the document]
JAicewizard: Could you elaborate a bit more?

While thinking about this I did find a couple cases where I can see the benefit of adding localstate, even if it complicated the API more. But I could not come up with a scenario where it matters how many threads are running.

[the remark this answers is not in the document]
JAicewizard: Yes I agree that a concurrency-aware `FillRow` would be needed, and that maybe adding locks to `FillRow` would be detrimental to the performance, that is indeed true. However if I add a localstate a lot of locking could be worked around, and functions may be able to use atomics instead of locks to avoid synchronization overhead.

I might implement a vector based API sometime this weekend, maybe that will help with this.

The vector vs row discussion is indeed a bigger than just this PR, which is important as users want more performance (by using vectorised APIs). I started with a row based API because it is probably conceptually the simplest for users.

JAicewizard:
I rebased and forcepushed to integrate datachunk changes. Also added a datachunk API for tablefunctions.

udf.go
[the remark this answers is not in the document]
JAicewizard: I added threaded variants of the options, where the threading is exposed and a functions to create local state is required. This removes any worry of threading for those that do not need it, but exposes the full API when someone wants to use it.

JAicewizard:
@taniabogatsch Do you have write access? if so could you run CI for this? I think I fixed all the issues in CI, but I would like to make sure.

Also do you know what the next steps are to getting this merged?

taniabogatsch:
@JAicewizard, I ran the CI. Regarding the next steps to getting this merged, ideally, we have the `DataChunk` support in place. But I'll have another look at the current state, with the changes from `main`, and maybe we can merge it sooner.

vector.go
```diff
@@ -75,6 +75,34 @@ func (vec *vector) tryCast(val any) (any, error) {
...
+		return convertNumericType[int64, R](v), nil
+	case uint:
+		return convertNumericType[uint, R](v), nil
+	case int:
+		return convertNumericType[int, R](v), nil
+	default:
+		return 0, nil
+	}
+
+}
```
taniabogatsch, line 104: Not used.

vector.go
taniabogatsch, line 1: We need to use the same functionality to set values here. Having two slightly different implementations for the same logic will be cumbersome. What prevents you from using `func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error` here? 

> The added code allows for setting values without having to turn them into an any. This significantly speeds up table functions, as converting a value into any does a heap allocation which is expensive.

Can we alter the existing implementation to improve the usage of `any`?

type.go
taniabogatsch, line 1: I would like to talk about a few things in this approach.

We should not use anything that keeps resources alive. The likeliness of unintentional memory leaks that are hard to find on the user's side is too high.

I don't mind introducing an exported type to hold (nested) duckdb types. They won't contain as much information as logical types, but we might be able to work around that. Then, we only need to create the logical type when we need it—and when we can immediately free it afterward.

If we discard the SQL approach (which, IMHO, is the most intuitive for users, but also ugly to implement bc parser), then we need to provide a very intuitive interface (also for nested types) for them to create these types. 

Maybe we can mix providing SQL names (`"LIST"`, `"INT"`, etc.) with functions like `SetListChildType`... 🤔 But I am a bit lost here w.r.t. what a good approach would be...

taniabogatsch (changes requested):
Hey! I just had another more detailed look at the PR. This is going in a great direction. 😄 I think we still need to iron out two main things before we add the 'finishing touches' to the PR.

1. Changes to `vector` and the custom functions there. See my comment on the respective file.
2. Discuss the `Type` approach.

Other remarks.
- Could you also run `gofumpt -l -w .` on your PR for more uniform formatting?
- It is nice to see the increased number of tests.
- Much improved documentation in `udtf.go` 👍

type.go
JAicewizard: > We should not use anything that keeps resources alive. The likeliness of unintentional memory leaks that are hard to find on the user's side is too high.

I do not think that this is currently the case, `reflect.Type` does not keep the value alive, but I do agree that we should not keep any values alive. This is why I chose for this API, since I found it very annoying to work with manually closing types (and the many ways to mis-use it).

>If we discard the SQL approach (which, IMHO, is the most intuitive for users, but also ugly to implement bc parser), then we need to provide a very intuitive interface (also for nested types) for them to create these types.
>
>Maybe we can mix providing SQL names ("LIST", "INT", etc.) with functions like SetListChildType... 🤔 But I am a bit lost here w.r.t. what a good approach would be...

I also agree here, but for this PR this is not needed. Maybe we could outsource these additions to a different PR?Maybe something like `NewListType(Type)` and `NewStructType([]Field)`, and similar constructors could be used to safely construct a type.

Sample implementation:

```
type (
    Uint8 struct {}
    Uint16 struct {}
    Uin32 struct {}
    Uint16 struct {}
    ....
    ListType struct {
	child Type
    }
    Type interface {
        ToDuckdb() C.duckdb_logical_type
    }
)

func NewListType(t Type){...}
func (lt ListType)ToDuckdb(){...}
```

This allows not only for a pure-go safe representation of duckdb types, it also allows users to implement their own types (For example types implemented by extensions). Although I am not sure the point is actually beneficial, we could keep `ToDuckdb` private-only, exposing it later is not a breaking change.

data_chunk.go
```diff
@@ -91,3 +107,11 @@ func (chunk *DataChunk) setSize() error {
...
 	return nil
 }
+
+func (chunk *DataChunk) SetSize(size int) {
+	C.duckdb_data_chunk_set_size(chunk.data, C.idx_t(size))
+}
+
+func (chunk *DataChunk) GetSize() int {
+	return int(C.duckdb_data_chunk_get_size(chunk.data))
+}
```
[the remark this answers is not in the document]
JAicewizard: All initializers set this value, just like `SetSize` assumes its set it seems safe

errors.go
```diff
@@ -44,15 +44,20 @@ func invalidatedAppenderError(err error) error {
...
-	invalidatedAppenderMsg = "appended data has been invalidated due to corrupt row"
+	driverErrMsg             = "database/sql/driver"
+	duckdbErrMsg             = "duckdb error"
+	castErrMsg               = "cast error"
+	structFieldErrMsg        = "invalid STRUCT field"
+	columnErrMsg             = "column index"
+	columnCountErrMsg        = "invalid column count"
+	unsupportedTypeErrMsg    = "unsupported data type"
+	invalidatedAppenderMsg   = "appended data has been invalidated due to corrupt row"
+	invalidTampleFunctionMsg = "table function was rejected by duckdb for unknown reason"
```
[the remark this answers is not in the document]
JAicewizard: I am not sure what you mean by this, I could make this an error using `errors.New` like the ones below this?

vector.go
JAicewizard: I changed the existing implementation to use the new generic functions, completely removing the old set functions and replacing them with the generic ones. I still need to implement a couple of types this way, but this was also my intention, just never got around to it. The changes made to the vector type made it much nicer to do.

> What prevents you from using func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error here?

The conversion to any would be expensive, but I removed the duplicate implementations so this should no longer be an issue

JAicewizard (commented):
I made some changes I have not pushed yet, ill push them soon.

JAicewizard:
Thanks for the feedback! My final comment got lost, but I pushed my changes already anyways. I force-pushed for the rebase and re-organisation of commits

JAicewizard:
No problem, I noticed some of your commits landing in duckdb as well :)

type_info.go
```diff
@@ -0,0 +1,414 @@
...
+			return nil, err
+		}
+		return NewMapInfo(kt, vt)
+	case reflect.Pointer, reflect.Interface:
+		return tryGetDuckdbTypeFromValue(rt.Elem())
+	// This case should never be reached
+	default:
+		return nil, unsupportedTypeError(rt.String())
+	}
+}
```
taniabogatsch, line 410: I'll leave this as a comment so that we can link to the section in https://github.com/marcboeker/go-duckdb/pull/272.

JAicewizard:
Rebased to the latest, and added tests for the parallel functions just like the normal ones.

I also made some small performance improvements WRT `Row`, and especially when using the chunk API it is very fast. The biggest bottleneck is importing the chunks from duckdb. Maybe in the future this can be sped up as well!

taniabogatsch:
I've deleted some duplicate comments and reviewed the PR again to align it with the scalar UDF PR.

Could you move the changes in the `vector*.go` files into a separate PR? My understanding is that they generally speed up go-duckdb by using fewer `any` types. That separate PR should then be fairly self-contained and quickly merged.

row.go
taniabogatsch, line 1: Should we move `Row` to `udf_utils.go`?

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+import "C"
+
+import (
+	"database/sql"
+	"runtime"
+	"runtime/cgo"
+	"unsafe"
+)
+
+type (
```
taniabogatsch, line 31: Let's move `ColumnMetaData`, `CardinalityData`, and `ThreadedTableSourceInitData` to `udf_utils.go`.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+	ColumnMetaData struct {
+		Name string // Name of the column
+		T    TypeInfo   // type of the column
+	}
+
+	// CardinalityData is used to indicate the cardinality of a (table)function.
+	// The IsExact is used to indicate wether or not the cardinality is exact.
+	// When it is impossible/difficult to determine the  cardinality, an inexact cardinality
+	// can be used.
+	CardinalityData struct {
```
taniabogatsch, line 43: Let's rename this to `CardinalityInfo`.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+		Cardinality uint // Cardinality
+		IsExact     bool // Wether or not the cardinality is exact
+	}
+
+	// ThreadedTableSourceInitData contains any info that can be passed to duckdb when
+	// initialising the tablefunction.
+	ThreadedTableSourceInitData struct {
```
taniabogatsch, line 50: Let's rename this to `ThreadInfo`.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+		T    TypeInfo   // type of the column
+	}
+
+	// CardinalityData is used to indicate the cardinality of a (table)function.
+	// The IsExact is used to indicate wether or not the cardinality is exact.
+	// When it is impossible/difficult to determine the  cardinality, an inexact cardinality
+	// can be used.
+	CardinalityData struct {
+		Cardinality uint // Cardinality
+		IsExact     bool // Wether or not the cardinality is exact
```
taniabogatsch, line 45: `Exact` instead of `IsExact`.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+	args := make([]any, argCount)
+	namedArgs := make(map[string]any)
+	for i, t := range config.Arguments {
+		value := C.duckdb_bind_get_parameter(info, C.idx_t(i))
+		var err error
+		args[i], err = getValue(t, value)
+		if err != nil {
+			errstr := C.CString(err.Error())
+			defer C.free(unsafe.Pointer(errstr))
+			C.duckdb_bind_set_error(info, errstr)
```
taniabogatsch, line 192: Let's move these occurrences into a function.
```Go
func setBindError(bind_info C.duckdb_bind_info, msg string) {
 	err := C.CString(msg)
        C.duckdb_bind_set_error(bind_info, err)
 	C.duckdb_free(unsafe.Pointer(err))
 }
```

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+
+//export udf_row_callback
+func udf_row_callback(info C.duckdb_function_info, output C.duckdb_data_chunk) {
+	h := *(*cgo.Handle)(C.duckdb_function_get_bind_data(info))
+	instance := h.Value().(pinnedValue[tableFunctionData]).value
+
+	var chunk DataChunk
+	// false, maybe this needs to be true but I don't think we need to read the validity mask
+	// It should be set after executing the function, but never read.
+	err := chunk.initFromDuckDataChunk(output, false)
```
taniabogatsch, line 301: It should be `true`; otherwise, setting values can cause uninitialized memory crashes.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+		C.duckdb_destroy_table_function(&tableFunction)
+		if state != 0 {
+			return invalidTableFunctionError()
+		}
+		return nil
+	})
+	return err
+}
+
+func getValue(t TypeInfo, v C.duckdb_value) (any, error) {
```
taniabogatsch, line 469: With the recent DuckDB release, getting significantly more values through the C API is now possible.
Should we move this into a separate file (`value.go` or similar) and extend the support for all types?
Otherwise, we can add a `FIXME` and open a follow-up PR.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+	instance := h.Value().(pinnedValue[tableFunctionData]).value
+
+	var chunk DataChunk
+	// false, maybe this needs to be true but I don't think we need to read the validity mask
+	// It should be set after executing the function, but never read.
+	err := chunk.initFromDuckDataChunk(output, false)
+	if err != nil {
+		errstr := C.CString(err.Error())
+		defer C.free(unsafe.Pointer(errstr))
+		C.duckdb_function_set_error(info, errstr)
```
taniabogatsch, line 305: Use `setFuncError` for all these occurrences.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+
+	columnCount := C.duckdb_init_get_column_count(info)
+	for i := C.idx_t(0); i < columnCount; i++ {
+		srcPos := int(C.duckdb_init_get_column_index(info, i))
+		instance.projection[srcPos] = int(i)
+	}
+
+	initData := instance.fun.(threadedTableSource).Init()
+	C.duckdb_init_set_max_threads(info, C.idx_t(initData.MaxThreads))
+}
```
taniabogatsch, line 278: Add a function for the following lines to avoid duplicate code.
```Go
	h := *(*cgo.Handle)(C.duckdb_init_get_bind_data(info))
 	instance := h.Value().(pinnedValue[tableFunctionData]).value

 	columnCount := C.duckdb_init_get_column_count(info)
 	for i := C.idx_t(0); i < columnCount; i++ {
 		srcPos := int(C.duckdb_init_get_column_index(info, i))
 		instance.projection[srcPos] = int(i)
 	}
```

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+		var x any = function
+		switch tfunc := x.(type) {
+		case RowTableFunction:
+			C.duckdb_table_function_set_bind(tableFunction, C.bind(C.udf_bind_row))
+			C.duckdb_table_function_set_function(tableFunction, C.callback(C.udf_row_callback))
+
+			config = tfunc.Config
+			if tfunc.BindArguments == nil {
+				return invalidTableFunctionError()
+			}
```
taniabogatsch, line 417: We can turn this into a more specific error message indicating what went wrong.

udtf.go
```diff
@@ -0,0 +1,493 @@
...
+			errstr := C.CString(err.Error())
+			defer C.free(unsafe.Pointer(errstr))
+			C.duckdb_function_set_error(info, errstr)
+		}
+	}
+}
+
+// RegisterTableUDF registers a TableFunctionProvider to duckdb.
+// Projectionpushdown is enabled by default, and implemented transparently.
+func RegisterTableUDF[TFT TableFunction](c *sql.Conn, name string, function TFT) error {
```
taniabogatsch, line 388: Let's return an error if the `name` is empty or the `function` is `nil`. Are there other invalid input scenarios?

taniabogatsch (changes requested):
I just looked at this PR again before merging the scalar UDF PR. :)
We should be able to reuse a few functions from there and finally finish up this PR.
My comments are primarily nits and naming suggestions to align the PR with DuckDB and go-duckdb naming conventions.

udtf.go
JAicewizard: I think this is fine for now, I will add a FIXME

row.go
```diff
@@ -0,0 +1,46 @@
...
+import "C"
+
+type (
+	// Row represents one row in duckdb. It references the vectors underneeth.
+	Row struct {
+		chunk      *DataChunk
+		r          C.idx_t
+		projection []int
+	}
+)
```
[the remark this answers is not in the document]
taniabogatsch: Yes, fair point. Let's keep the file and name separate.

row.go
taniabogatsch: I agree with the non-descriptive name but didn't have a better idea yesterday. Let's keep the `row.go` file. We can rename `udf_utils.go` or split it into `pinner.go` and `udf.go` (with functions for both scalar and table UDFs).

udtf.go
JAicewizard: function cannot be nil, but I did add empty name

udtf.go
taniabogatsch: Yes, I can see giving this plan a shot.
I am just a bit conflicted about the name `table_source.go`, since a source emits data, and appending data to a table sounds like the contrary. 🤔 
What about something like `chunk_manager`? We never really talk about the table, only data chunks.

row.go
JAicewizard: Oh I didn't know you alread had that file haha, I moved some pinner stuff into there as well. I think its fine now, maybe rename it to udf_common but that is insignificant. I think everything in there is too small and insignificant to warrant its own thing so lets just leave it. Unless we start using the pinner in other places it may be worth moving it out, but its just a utility.

row.go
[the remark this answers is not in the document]
JAicewizard: Should I move this into a new file now? or only when the new appender gets implemented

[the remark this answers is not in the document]
taniabogatsch: Let's keep this file for now. 👍

udtf.go
JAicewizard: Hmm, I think table source would work since you are essentially taking the union of 2 tables, the appending data is just the specific implementation of how that is done.

> since a source emits data, and appending data to a table sounds like the contrary

True, the way I thought about is, is that you have this source (user), and a sink (duckdb), and go-duckdb is just the connector. But I see your point. I will think about it a bit more.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+	"log"
+	"reflect"
+
+	"github.com/marcboeker/go-duckdb"
+)
+
+var db *sql.DB
+
+type tableUDF struct {
+	n     int64
```
taniabogatsch, line 16: Maybe we can make the name for `n` more descriptive, as this will be the main example for people when using table functions.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+
+func BindTableUDF(namedArgs map[string]any, args ...interface{}) (duckdb.RowTableSource, error) {
+	return &tableUDF{
+		count: 0,
+		n:     args[0].(int64),
+	}, nil
+}
+
+func (d *tableUDF) Columns() []duckdb.ColumnInfo {
+	t, _ := duckdb.NewTypeInfo(duckdb.TYPE_BIGINT)
```
taniabogatsch, line 28: Let's add an error check: `check(err)`.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+func main() {
+	var err error
+	db, err = sql.Open("duckdb", "?access_mode=READ_WRITE")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+	conn, _ := db.Conn(context.Background())
+
+	t, _ := duckdb.NewTypeInfo(duckdb.TYPE_BIGINT)
```
taniabogatsch, line 61: `check(err)`

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+	fun := duckdb.RowTableFunction{
+		Config: duckdb.TableFunctionConfig{
+			Arguments: []duckdb.TypeInfo{t},
+		},
+		BindArguments: BindTableUDF,
+	}
+
+	duckdb.RegisterTableUDF(conn, "whoo", fun)
+
+	check(db.Ping())
```
taniabogatsch, line 71: Do you think we should remove this? It doesn't add anything to the example.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+		Exact:     true,
+	}
+}
+
+func main() {
+	var err error
+	db, err = sql.Open("duckdb", "?access_mode=READ_WRITE")
+	if err != nil {
+		log.Fatal(err)
+	}
```
taniabogatsch, line 57: Let's consistently use `check(err)` in the example.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+			case nil:
+				fmt.Print(columns[i], ": NULL")
+			case []byte:
+				fmt.Print(columns[i], ": ", string(value.([]byte)))
+			default:
+				fmt.Print(columns[i], ": ", value)
+			}
+			fmt.Printf("\nType: %s\n", reflect.TypeOf(value))
+		}
+		fmt.Println("-----------------------------------")
```
taniabogatsch, line 106: Maybe we could add some result verification here. Also, there are no `nil` or `[]byte` values, so we can just assume `BIGINT` and remove the switch.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+
+	t, _ := duckdb.NewTypeInfo(duckdb.TYPE_BIGINT)
+	fun := duckdb.RowTableFunction{
+		Config: duckdb.TableFunctionConfig{
+			Arguments: []duckdb.TypeInfo{t},
+		},
+		BindArguments: BindTableUDF,
+	}
+
+	duckdb.RegisterTableUDF(conn, "whoo", fun)
```
taniabogatsch, line 69: Haha, `whoo`! 😄 
However, we should choose a more descriptive name to make it easier for people to follow along with the example.

examples/table_udf/main.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"log"
+	"reflect"
+
+	"github.com/marcboeker/go-duckdb"
+)
+
+var db *sql.DB
+
+type tableUDF struct {
```
taniabogatsch, line 15: Should we use a more descriptive name of what the function does?

examples/table_udf/main.go
taniabogatsch, line 1: Since there are several different ways to use table functions, it would be helpful to extend the examples slightly.
So people can use the table functions smoothly. :)

Maybe we can call a few different functions from the `main` function, and each shows an example of a different table function in action? E.g., row-based, chunk-based, threaded?

udf_utils.go
```diff
@@ -24,12 +28,28 @@ func (v pinnedValue[T]) unpin() {
 	v.pinner.Unpin()
 }
 
+func getPinnedValueValue[T any](handle unsafe.Pointer) T {
```
taniabogatsch, line 31: Maybe `getPinned` is enough, and we do not have to decide between `ValueValue` and `Value`. 😄

taniabogatsch (commented):
I left a few comments about the included example.
Beyond that, I'll try to review the changes in `udtf.go` and `udtf_test.go` tomorrow or later this week.

JAicewizard:
Thanks for the PR, I added a parallel example. While doing so I realised the parallel test was completely wrong and parallel functions did not work at all, so I fixed those too.

udtf.go
```diff
@@ -0,0 +1,446 @@
...
+	maxSize := C.duckdb_vector_size()
+
+	switch fun := instance.fun.(type) {
+	case RowTableSource:
+		// At the end of the loop row.r must be the index one past the last added row
+		for row.r = 0; row.r < maxSize; row.r++ {
+			nextResults, err := fun.FillRow(row)
+			if err != nil {
+				setFuncError(info, err.Error())
+			}
```
taniabogatsch, line 303: Should we abort if we encounter an error?
Same for the `ThreadedRowTableSource` case.

examples/table_udf_basic/main.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+func main() {
+	var err error
+	db, err = sql.Open("duckdb", "?access_mode=READ_WRITE")
+	check(err)
+	if err != nil {
+		log.Fatal(err)
+	}
```
taniabogatsch, line 60: Remove.

examples/table_udf_parallel/main.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+func main() {
+	var err error
+	db, err = sql.Open("duckdb", "?access_mode=READ_WRITE")
+	check(err)
+	if err != nil {
+		log.Fatal(err)
+	}
```
taniabogatsch, line 112: Remove.

taniabogatsch (commented):
I've left more comments, but this is probably almost done. Most of my comments are about test coverage. 
It's good to hear that you found and fixed that bug in the parallel execution. 😄

JAicewizard:
I started adding more tests, and I found one really odd error. On the latest commit, if you switch around the arguments to `SetValue` for the chunk test, it returns an error as it should. However, then later duckdb creates a SIGSEGV and I don't really know why. Somehow the chunk seems to be corrupt or something? However I don't see how that could happen

udtf_test.go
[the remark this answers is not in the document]
JAicewizard: Yup did all of this, also added more types. The usage of the timestamp times are a bit wonky, but I figured it out with the tests

udtf.go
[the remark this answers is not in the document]
JAicewizard: I renamed them to tableUDF.go for more idiomatic naming

JAicewizard (commented):
I forgot to press the submit button last time, so there are 6 comments of which I don't know how relevant thy still are. But I fixed all the issues you pointed out

taniabogatsch:
Hi @JAicewizard - thanks for the changes! If you don't mind, then I'll go ahead, run some formatting and spell checking, and some minor renaming, and nits. And merge it afterwards?

taniabogatsch:
> I started adding more tests, and I found one really odd error. On the latest commit, if you switch around the arguments to SetValue for the chunk test, it returns an error as it should. However, then later duckdb creates a SIGSEGV and I don't really know why. Somehow the chunk seems to be corrupt or something? However I don't see how that could happen

I'll also have a look at this 👍

taniabogatsch:
https://github.com/JAicewizard/go-duckdb/pull/2

taniabogatsch (approved):

taniabogatsch:
@JAicewizard, many thanks for all your work on this! 🥳

JAicewizard:
It was a long road, but a lot of things had to happen for this to be mergable :)

taniabogatsch:
Haha yes! :D

PR #205, opened 2024-05-01
Use build tags for Arrow implementation

### LINKED ISSUE #204: Make Apache Arrow Optional
#134 added support for Apache Arrow but introduces a pretty heavy dependency in the process. Analysing my binaries using [`gsa`](https://github.com/Zxilly/go-size-analyzer), `github.com/apache/arrow/go/v14` introduced an additional ~800kb to my binary alone. In fact, that dependency also imports `github.com/goccy/go-json` which introduces an additional ~750kb which seems excessive if you do not use Arrow.

Would it be possible to introduce build tags and make this feature optional instead?

ayuhito:
Closes #204.

I tested the build flags on my project and saw around ~2MB shaved off the final binary, which is a solid success. Including the `duckdb_arrow` build tag reverted it back to the old size, so it seems that the tag is working.

I'm not too sure what the breaking change policy is for this library, so I'm happy to make appropriate changes. However, I strongly believe features like this should be opt-in, rather than opt-out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ examples:
 
 .PHONY: test
 test:
-	go test -v -race -count=1 -tags="duckdb_arrow" .
+	go test -v -race -count=1 .
 
 .PHONY: deps.header
 deps.header:
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ examples:
 
 .PHONY: test
 test:
-	go test -v -race -count=1 .
+	go test -v -race -count=1 -tags="duckdb_arrow" .
 
 .PHONY: deps.header
 deps.header:
--- a/README.md
+++ b/README.md
@@ -131,13 +131,7 @@ if err != nil {
 
 ## DuckDB Apache Arrow Interface
 
-To use the [DuckDB Arrow Interface](https://duckdb.org/docs/api/c/api#arrow-interface), you need to build your binary with the `duckdb_arrow` build tag.
-
-```sh
-go build -tags="duckdb_arrow"
-```
-
-Then, you can obtain a new `Arrow` by passing a DuckDB connection to `NewArrowFromConn()`.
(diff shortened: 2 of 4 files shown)


taniabogatsch (commented):
The changes look neat, and I agree that opt-in is sensible for arrow support.

taniabogatsch:
Hi @ayuhito, I've discussed this with @marcboeker. While we agree that opt-in is more sensible for features like the arrow support, the appender, and upcoming features like UDFs, we do not want to break the current version. Can this PR be changed to support easy opt-out for these and keep the default build as-is? Then, once go-duckdb moves to a v2, we can swap to opt-in. What do you think?

ayuhito:
Hi @taniabogatsch. That makes perfect sense. I'll update the PR to use `no_duckdb_arrow` instead.

For clarity, I'm not sure if offering build tag options for the appender and upcoming UDF features will make a noticeable difference to the code since they are small snippets that do not introduce any new dependencies. I'll verify if omitting `appender.go` makes a difference, but I'm suspecting it would be unnecessary.

Also regarding V2, may I suggest using [a similar pattern](https://github.com/marcboeker/go-duckdb/issues/204#issuecomment-2083323061) mentioned in the edit (although it being a slightly more annoying migration path for users). It eliminates the need for build tags and enables proper deadcode elimination to take place from the Go linker side. It also handles the appender and UDF cases too, however small the gains may be.

taniabogatsch:
Great, thanks! Let's focus on opt-out for `arrow` in this PR, then.

Your point w.r.t. the minimal impact for the appender, etc. is fair. With that in mind, making them opt-out (and eventually opt-in) would only create a logical separation between driver functionality (`database/sql`) and additional features. 🤔 Which will be much cleaner by introducing [your pattern suggestion](https://github.com/marcboeker/go-duckdb/issues/204#issuecomment-2083323061).

taniabogatsch (approved):

ayuhito:
Conflicts should be resolved now. 👍 Didn't realize the GitHub Actions workflows that I was tinkering with in a separate branch will affect this PR.

PR #208, opened 2024-05-07
Replacement scan

ajzo90:
Simple implementation of replacement scan API in duckdb. 

Registers a callback function that gets called when a query refer to a table that not exists in duckdb as per https://duckdb.org/docs/api/c/replacement_scans.

Currently limited to supporting int64 and string as table function arguments, but this seems like a limitation in duckdb C API.

### DIFF
--- a/replacement_scan.go
+++ b/replacement_scan.go
@@ -17,18 +17,19 @@ type ReplacementScanCallback func(tableName string) (string, []any, error)
 
 func RegisterReplacementScan(connector *Connector, cb ReplacementScanCallback) {
 	handle := cgo.NewHandle(cb)
-	C.duckdb_add_replacement_scan(connector.db, C.duckdb_replacement_callback_t(C.replacement_scan_cb), unsafe.Pointer(handle), C.duckdb_delete_callback_t(C.replacement_scan_destroy_data))
+	C.duckdb_add_replacement_scan(connector.db, C.duckdb_replacement_callback_t(C.replacement_scan_cb), unsafe.Pointer(&handle), C.duckdb_delete_callback_t(C.replacement_scan_destroy_data))
 }
 
 //export replacement_scan_destroy_data
 func replacement_scan_destroy_data(data unsafe.Pointer) {
-	h := cgo.Handle(data)
+	h := *(*cgo.Handle)(data)
 	h.Delete()
 }
 
 //export replacement_scan_cb
 func replacement_scan_cb(info C.duckdb_replacement_scan_info, table_name *C.cchar_t, data *C.void) {
-	scanner := cgo.Handle(unsafe.Pointer(data)).Value().(ReplacementScanCallback)
+	h := *(*cgo.Handle)(unsafe.Pointer(data))
+	scanner := h.Value().(ReplacementScanCallback)
 	tFunc, params, err := scanner(C.GoString(table_name))
 	if err != nil {
 		errstr := C.CString(err.Error())
--- a/replacement_scan.go
+++ b/replacement_scan.go
@@ -19,12 +19,11 @@ var replacementScanState struct {
 	callback ReplacementScanCallback
 }
 
-func RegisterReplacementScan(connector *Connector, cb ReplacementScanCallback) error {
+func RegisterReplacementScan(connector *Connector, cb ReplacementScanCallback) {
 	replacementScanState.mtx.Lock()
 	replacementScanState.callback = cb
 	replacementScanState.mtx.Unlock()
 	C.duckdb_add_replacement_scan(connector.db, C.duckdb_replacement_callback_t(C.replacement_scan_cb), nil, C.duckdb_delete_callback_t(C.free))
-	return nil
 }
 
 //export replacement_scan_cb
--- a/replacement_scan.go
+++ b/replacement_scan.go
@@ -8,29 +8,23 @@ package duckdb
 */
 import "C"
 import (
-	"sync"
 	"unsafe"
 )
 
 type ReplacementScanCallback func(tableName string) (string, []any, error)
 
 var replacementScanState struct {
-	mtx      sync.Mutex
 	callback ReplacementScanCallback
 }
 
 func RegisterReplacementScan(connector *Connector, cb ReplacementScanCallback) {
-	replacementScanState.mtx.Lock()
 	replacementScanState.callback = cb
-	replacementScanState.mtx.Unlock()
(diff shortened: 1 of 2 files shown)


replacement_scan.go
```diff
@@ -0,0 +1,65 @@
+package duckdb
+
+/*
+   #include <stdlib.h>
+   #include <duckdb.h>
+   void replacement_scan_cb(duckdb_replacement_scan_info info, const char *table_name, void *data);
+   typedef const char cchar_t;
```
taniabogatsch, line 7: Using `*C.char` directly does not work?

replacement_scan.go
```diff
@@ -0,0 +1,65 @@
...
+import "C"
+import (
+	"sync"
+	"unsafe"
+)
+
+type ReplacementScanCallback func(tableName string) (string, []any, error)
+
+var replacementScanFnMxt sync.Mutex
+var replacementScanCallback ReplacementScanCallback
```
taniabogatsch, line 18: Maybe it is sensible to move this into a struct instead of individual variables for extensibility? Then, there can be a global `var` of that struct?

replacement_scan.go
```diff
@@ -0,0 +1,65 @@
...
+	replacementScanFnMxt.Unlock()
+	C.duckdb_add_replacement_scan(connector.db, C.duckdb_replacement_callback_t(C.replacement_scan_cb), nil, C.duckdb_delete_callback_t(C.free))
+	return nil
+}
+
+//export replacement_scan_cb
+func replacement_scan_cb(info C.duckdb_replacement_scan_info, table_name *C.cchar_t, data *C.void) {
+	replacementScanFnMxt.Lock()
+	scanner := replacementScanCallback
+	replacementScanFnMxt.Unlock()
```
taniabogatsch, line 32: To clarify, this synchronises replacing the currently registered replacement scan with a new one?

replacement_scan.go
```diff
@@ -0,0 +1,65 @@
...
+
+var replacementScanFnMxt sync.Mutex
+var replacementScanCallback ReplacementScanCallback
+
+func RegisterReplacementScan(connector *Connector, cb ReplacementScanCallback) error {
+	replacementScanFnMxt.Lock()
+	replacementScanCallback = cb
+	replacementScanFnMxt.Unlock()
+	C.duckdb_add_replacement_scan(connector.db, C.duckdb_replacement_callback_t(C.replacement_scan_cb), nil, C.duckdb_delete_callback_t(C.free))
+	return nil
```
taniabogatsch, line 25: Do we need to return a `nil` error here? Is this for future extensibility?

taniabogatsch (changes requested):
Hi @ajzo90, thanks for the PR! I've added a few comments and questions.

replacement_scan.go
ajzo90: Yeah, I remember this was a bit tricky. Perhaps there are better alternatives.

replacement_scan.go
ajzo90: Yeah, let's remove it.

replacement_scan.go
ajzo90: That was my thought. It gives the caller more control over how to deal with errors, on the other hand it as annoying if concurrent libraries introduces it's own replacement. 
I think it is better to let the client decide how to deal with concurrent use of replacement.

replacement_scan_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	rows, err := db.Query("select * from any_table")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		rangeRows--
+	}
```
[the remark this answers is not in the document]
ajzo90: My thought was just to make sure it triggered the range function, for me this is enough evidence for that. Perhaps we should add other test cases. It's easier to construct them when the tUDF is merged :)

replacement_scan.go
taniabogatsch: Okay, no worries.

replacement_scan_test.go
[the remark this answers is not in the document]
taniabogatsch: Fair point.

replacement_scan.go
taniabogatsch: Yes, I agree. Moving this to the application level makes more sense to me... So could we take it out of this PR? I would still keep the `var` wrapper, even if the struct then only contains a single element.

replacement_scan.go
```diff
@@ -0,0 +1,66 @@
...
+	defer C.free(unsafe.Pointer(fNameStr))
+
+	for _, v := range params {
+		switch x := v.(type) {
+		case string:
+			str := C.CString(x)
+			val := C.duckdb_create_varchar(str)
+			C.duckdb_replacement_scan_add_parameter(info, val)
+			C.free(unsafe.Pointer(str))
+			C.free(unsafe.Pointer(val))
```
taniabogatsch, line 54: This should be `C.duckdb_destroy_value(&val)`, I think...

replacement_scan_test.go
taniabogatsch, line 1: Let's also replace the remaining two `t.Fatal` with `require...`.

taniabogatsch (changes requested):
I had another look, and it looks almost ready to go in from my side. :) I've just added a few remaining comments.

replacement_scan.go
ajzo90: Right, nice catch!

replacement_scan.go
```diff
@@ -0,0 +1,60 @@
...
+   typedef const char cchar_t;
+*/
+import "C"
+import (
+	"unsafe"
+)
+
+type ReplacementScanCallback func(tableName string) (string, []any, error)
+
+var replacementScanState struct {
```
marcboeker, line 16: Is there probably another way to inject the callback into the `replacement_scan_cb` function instead of declaring it on the package level? What happens when opening multiple DBs and registering a `ReplacementScanCallback` for each of them? Will there be an interference between the multiple declarations, because `replacementScanState` will be overwritten?

ajzo90: Good point. Didn't think about such a scenario, luckily it was possible to store the callback inside the duckdb instance.

marcboeker:
@ajzo90 Thanks for the contribution.
@taniabogatsch Thanks for the review.

PR #213, opened 2024-05-13
Move to macos-latest

carlopi:
This seems to solve the issues in testing.

I am not sure why the problem is there across different versions, and if that is worth checking.


marcboeker:
@carlopi Thanks!

PR #222, opened 2024-05-31
[Feature] Scalar UDF support

taniabogatsch:
This is a draft PR. I built with [duckdb's feature branch](https://github.com/duckdb/duckdb/tree/feature), as scalar UDFs are not yet part of the released C API. The test failures are related to that.

Because of that, and because this PR depends on #219, there are a lot of file changes in this PR, as well as separate libraries. They can all be ignored. The relevant files are listed below.
- `scalar_udf.go`
- `scalar_udf_test.go`

@JAicewizard, I solved passing types differently than you did in #201. What do you think? Is passing SQL type names a sensitive idea? I also tried to solely use the `DataChunk` API draft's functions.

In later PRs, we can extend this with `ExecuteColumn`.

Here is the example included in this PR.
```go
type scalarUDF struct {
	err error
}

func (udf *scalarUDF) Config() ScalarFunctionConfig {
	return ScalarFunctionConfig{
		InputTypes: []string{"INT", "INT"},
		ResultType: "INT",
	}
}

func (udf *scalarUDF) ExecuteRow(args []driver.Value) (any, error) {
	if len(args) != 2 {
		return nil, errors.New("expected two values")
	}
	val := args[0].(int32) + args[1].(int32)
	return val, nil
}

func (udf *scalarUDF) SetError(err error) {
	udf.err = err
}

func TestScalarUDFPrimitive(t *testing.T) {
	require.NoError(t, err)

	c, err := db.Conn(context.Background())
	require.NoError(t, err)

	var udf scalarUDF
	err = RegisterScalarUDF(c, "my_sum", &udf)

	var msg int
	row := db.QueryRow(`SELECT my_sum(10, 42) AS msg`)
	require.NoError(t, row.Scan(&msg))
	require.Equal(t, 52, msg)
	require.NoError(t, db.Close())
}
```


JAicewizard:
> @JAicewizard, I solved passing types differently than you did in https://github.com/marcboeker/go-duckdb/pull/201. What do you think? Is passing SQL type names a sensitive idea?

I like the idea, however I think it might be difficult to handle composite types. Handling these would require a parser for to go from "struct{x:INT, y:DOUBLE}" to a duckdb type. Personally I also prefer to have any errors be returned when "creating" the types, since then creation of the error is very close to where it problem lies. Another thing to think about is what to do with aliases that duckdb uses, do we want to mirror those, and if we do how will we keep the list updated? (and do we change the name when duckdb does?)

JAicewizard:
On a seperate note, I like how few methods you need to implement, I think the tablefunctions are very complex to implement. This is kind of inherent to the type of function, but I still dislike it

scalar_udf.go
```diff
@@ -0,0 +1,152 @@
...
+	Config() ScalarFunctionConfig
+	ExecuteRow(args []driver.Value) (any, error)
+	SetError(err error)
+}
+
+//export scalar_udf_callback
+func scalar_udf_callback(info C.duckdb_function_info, input C.duckdb_data_chunk, output C.duckdb_vector) {
+	// info is a void* pointer to our ScalarFunction.
+	h := *(*cgo.Handle)(unsafe.Pointer(info))
+	scalarFunction := h.Value().(ScalarFunction)
```
JAicewizard, line 44: This is invalid use of `cgo.Handle`, and it bit me hard last week. `cgo.Handle` is already a pointer, no need to wrap another pointer around it. In fact you can sometimes get cgo errors when using it with a pointer like this, which really confused me at first.
You probably copied this from me so just letting you know before you get any issues with this :)

taniabogatsch: I copied it from their (Go) documentation example, but I am confused, so I am no saying it's not wrong, haha. I figured the second example applies, as `info` is a `void *` pointer?
```C
//! Additional function info. When setting this info, it is necessary to pass a destroy-callback function.
typedef struct _duckdb_function_info {
	void *__val;
} * duckdb_function_info;
```

```Go
// Handle provides a way to pass values that contain Go pointers
// (pointers to memory allocated by Go) between Go and C without
// breaking the cgo pointer passing rules. A Handle is an integer
// value that can represent any Go value. A Handle can be passed
// through C and back to Go, and Go code can use the Handle to
// retrieve the original Go value.
//
// The underlying type of Handle is guaranteed to fit in an integer type
// that is large enough to hold the bit pattern of any pointer. The zero
// value of a Handle is not valid, and thus is safe to use as a sentinel
// in C APIs.
//
// For instance, on the Go side:
//
//	package main
//
//	/*
//	#include <stdint.h> // for uintptr_t
//
//	extern void MyGoPrint(uintptr_t handle);
//	void myprint(uintptr_t handle);
//	*/
//	import "C"
//	import "runtime/cgo"
//
//	//export MyGoPrint
//	func MyGoPrint(handle C.uintptr_t) {
//		h := cgo.Handle(handle)
//		val := h.Value().(string)
//		println(val)
//		h.Delete()
//	}
//
//	func main() {
//		val := "hello Go"
//		C.myprint(C.uintptr_t(cgo.NewHandle(val)))
//		// Output: hello Go
//	}
//
// and on the C side:
//
//	#include <stdint.h> // for uintptr_t
//
//	// A Go function
//	extern void MyGoPrint(uintptr_t handle);
//
//	// A C function
//	void myprint(uintptr_t handle) {
//	    MyGoPrint(handle);
//	}
//
// Some C functions accept a void* argument that points to an arbitrary
// data value supplied by the caller. It is not safe to coerce a [cgo.Handle]
// (an integer) to a Go [unsafe.Pointer], but instead we can pass the address
// of the cgo.Handle to the void* parameter, as in this variant of the
// previous example:
//
//	package main
//
//	/*
//	extern void MyGoPrint(void *context);
//	static inline void myprint(void *context) {
//	    MyGoPrint(context);
//	}
//	*/
//	import "C"
//	import (
//		"runtime/cgo"
//		"unsafe"
//	)
//
//	//export MyGoPrint
//	func MyGoPrint(context unsafe.Pointer) {
//		h := *(*cgo.Handle)(context)
//		val := h.Value().(string)
//		println(val)
//		h.Delete()
//	}
//
//	func main() {
//		val := "hello Go"
//		h := cgo.NewHandle(val)
//		C.myprint(unsafe.Pointer(&h))
//		// Output: hello Go
//	}
type Handle uintptr
```

taniabogatsch:
> On a seperate note, I like how few methods you need to implement, I think the tablefunctions are very complex to implement. This is kind of inherent to the type of function, but I still dislike it

Yes, your PR was a great help to write this! And I noticed the same, there is significantly more logic around the table UDFs.

> I like the idea, however I think it might be difficult to handle composite types. Handling these would require a parser for to go from "struct{x:INT, y:DOUBLE}" to a duckdb type. 

> Another thing to think about is what to do with aliases that duckdb uses, do we want to mirror those, and if we do how will we keep the list updated? (and do we change the name when duckdb does?)

The parser is a fair argument against this. We currently have `func logicalTypeName(lt C.duckdb_logical_type) string {...}`, which does a bit of that, but it would have to be much improved. Additionally, the second argument is a strong one for me. If we stick to the go types and go-duckdb types, then they match the casts from `any` in the execution function. Maybe we can draft a `Type` interface in a separate PR, though? As it is a general concept that we want to introduce for UDFs... And it should integrate well with the existing exposed types... 🤔 

> Personally I also prefer to have any errors be returned when "creating" the types, since then creation of the error is very close to where it problem lies.

This is independent of which strategy we decide on for the types, no? We can return an error when registering the UDF, if we detect invalid types.

scalar_udf.go
JAicewizard: huh, that is a good point. I will revert this in my PR and try again but I got some errors. Maybe I should open a bug report for go? I definitely got `...invalid handle...` errors when originally, but I will see if I can find out whats wrong.

JAicewizard:
> This is independent of which strategy we decide on for the types, no? We can return an error when registering the UDF, if we detect invalid types.

true, but for, for example, table UDFs, the types returned types are only determined at bind, it would turn into a generic sql error. But this is definitely debatable, its just my opinion. A seperate PR could be appropriate, I don't have much time due to exams next week. Feel free to copy my implementation if you want, also keep in mind we can add new "constructors" whenever we want, if some duckdb API comes out to parse a string into a type we can do this. (we _can_ even make type an interface)

scalar_udf.go
JAicewizard: See https://github.com/golang/go/issues/68044 for what was happening. We should pin(and unpin) the handle before handing it off and after deleting it

taniabogatsch: Thanks a lot for diving into this and finding the proper approach!

taniabogatsch:
@JAicewizard, somehow this PR broke along the way.
The pointer I now receive in the callback does not longer contain the correct handle.
https://github.com/marcboeker/go-duckdb/pull/222/commits/f1d079a13ec38104380dd2e177d715f212be3e41

I also tried pinning (with `runtime.Pinner`), without success. Maybe I am missing something obvious? Any help would be appreciated.

taniabogatsch:
Never mind, it is most likely a bug introduced somewhere in the C API, as reverting to an older duckdb build (seems) to fix it. I.e., the same code works with `b63142c`.

JAicewizard:
Haha Yeah debugging these kind of issues it not fun. Nice that it is fixed. I will look for similar changes in duckdb for the table UDFs when rebasing my branch.

taniabogatsch:
Open TODOs:
- [ ] Add support for scalar function sets.
- [ ] Add support for variadic input parameters.
- [ ] Add support for `ANY` input parameters.

JAicewizard:
Followup from the types PR:
What if we make the `Config` method not return an error? Then the user would either need to handle the errors before passing the functions to go-duckdb (and thus compute these types beforehand as well), or ignore the errors (and thus be really confident about their types). If you disagree with this then I will just implement it with an error as well, but conceptually I think it shouldn't have to.

taniabogatsch:
> Followup from the types PR:
What if we make the Config method not return an error? Then the user would either need to handle the errors before passing the functions to go-duckdb (and thus compute these types beforehand as well), or ignore the errors (and thus be really confident about their types). If you disagree with this then I will just implement it with an error as well, but conceptually I think it shouldn't have to.

This sounds sensible to me. The user can either ignore the error or panic. And otherwise, we can throw in `RegisterFunction...`, which returns an `error`. I'll finish the Type interface PR and then push the changes here as well.

vector.go
```diff
@@ -28,6 +28,9 @@ type vector struct {
 	// The child vectors of nested data types.
 	childVectors []vector
 
+	// FIXME: this is a workaround until the C API exposes SQLNULL.
+	isSQLNull bool
```
JAicewizard, line 32: It would be nice to have some more documentation on what this actually does.

Does this also impact table functions? if so it would be nice to know so I can use it too

scalar_udf.go
```diff
@@ -0,0 +1,241 @@
...
+	"database/sql"
+	"database/sql/driver"
+	"runtime/cgo"
+	"unsafe"
+)
+
+type ScalarFuncConfig interface {
+	InputTypeInfos() []TypeInfo
+	ResultTypeInfo() TypeInfo
+}
```
JAicewizard, line 25: Would this not be better as a struct? If in the future we would need to add some method to this it would be a breaking change, adding a struct field wouldn't have to be as breaking.

scalar_udf.go
```diff
@@ -0,0 +1,241 @@
...
+		C.duckdb_destroy_scalar_function(&scalarFunc)
+		if state == C.DuckDBError {
+			return getError(errAPI, errScalarUDFCreate)
+		}
+		return nil
+	})
+	return err
+}
+
+func RegisterScalarUDFSet(c *sql.Conn, name string, functions ...ScalarFunc) error {
```
JAicewizard, line 208: Although I like the API ATM, I think this will cause issues when adding chunked and threaded functions (I assume those also exist for scalar functions). How would you combine these into a single UDF set? This API could be changed to allow `... any` as paramter but that would also be bad.
I think it would be more future-proof to make the user first create the set, then add them one by one, and then register to duckdb. This way a different function can be used for adding chunk scalar functions and row scalar functions (using generics or not, doesn't matter).

JAicewizard (commented):
This looks very nice, it looks a bit messy with all the type_info changes in here as well, but I think this looks good to be merged if the comments are addressed.

Optionally you could add parallel and chunk APIs as well, but I don't think it has any unfixable implications besides what I already mentioned.

vector.go
taniabogatsch: The input parameter can have no type, for example: `SQLNULL`, `LIST(SQLNULL)`, `STRUCT(VARCHAR: SQLNULL)`.
```sql
SELECT my_func(NULL, [NULL], {'hello': NULL});
```

So, `isSQLNull` is set when inferring the type from a DuckDB data chunk. Currently, we see `INVALID` as the type because the C API does not (yet) expose `SQLNULL`: https://github.com/duckdb/duckdb/pull/13999.

This is only relevant for scenarios where untyped input is possible. In the value getter and setter functions, we treat `SQLNULL` as `nil`.

taniabogatsch: I've elaborated on the `FIXME`.

scalar_udf.go
taniabogatsch: That's a good catch; I'll add the changes.
Also, I'll add a chunked row execution example + interface similar to your PR.

scalar_udf.go
taniabogatsch: That's a good point. 👍

taniabogatsch:
Thanks for your review!

> Optionally you could add parallel and chunk APIs as well, but I don't think it has any unfixable implications besides what I already mentioned.

I am unsure what you mean with the parallel API, I need to check with your table UDF PR again.
I'll add the chunk API 👍 Even though we are still missing helper functions like `SetColumn`.

taniabogatsch:
I also merged the type interface changes in `main,` so hopefully, it is less messy now.

vector.go
JAicewizard: Ah so it is a type indicating the entire column is `NULL`? That sounds nice to know indeed.

JAicewizard:
> I am unsure what you mean with the parallel API, I need to check with your table UDF PR again.

I cannot find the scalar function API online at all, so I can't check ATM, but table functions can specify their max threads that they can execute on, and executing on more than one thread of course requires being aware of this and handling local vs global state (thus I implemented them as different types). I don't know if something similar exists for scalar functions

>  I'll add the chunk API 👍 Even though we are still missing helper functions like SetColumn.

Ah yeah I see the problem, I currently use `chunk.SetValue(i, 0, d.count)` in the chunk variant, so I still set the individual values. This is surprisingly fast, so I don't see a reason we would need a `SetColumn` right now, although it is of course good to implement this in the future.

vector.go
taniabogatsch: Exactly. 👍

taniabogatsch:
> This is surprisingly fast, so I don't see a reason we would need a SetColumn right now, although it is of course good to implement this in the future.

Yes, I achieved speed-ups on this with https://github.com/marcboeker/go-duckdb/pull/254.
We hold the memory pointers in the `vector`, so `SetValue` performers direct access to them without too much function call overhead, etc. It is not quite at the level of vectorized execution over the chunk, but I am also happy with the performance.

taniabogatsch:
> I cannot find the scalar function API online at all, so I can't check ATM, but table functions can specify their max threads that they can execute on, and executing on more than one thread of course requires being aware of this and handling local vs global state (thus I implemented them as different types). I don't know if something similar exists for scalar functions

Ah yes, I've been working with `duckdb.h` directly, which contains the functions.
I opened an issue here to add the scalar functions to the documentation: https://github.com/duckdb/duckdb-web/issues/3679.

JAicewizard:
The table function documentation is also broken ATM, I will open an issue about that too.

But I just realised that scalar functions probably don't need any state at all, so there probably is no max threads for scalar functions.

scalar_udf.go
taniabogatsch: For scalar functions, we always use chunks in the execution callback. We mirror duckdb's internal chunked execution engine here, so there are no different scalar function types, as is the case with table functions.

I've changed the API to allow adding more callback function types without creating a breaking change.
If, at any point, we want to expose `type chunkFn func(chunk DataChunk) (any, error)`, we can do so, but it might not become necessary. What do you think?
```Go
type rowFn func(args []driver.Value) (any, error)

type ScalarFuncExecutor struct {
	RowExecutor rowFn
}

type ScalarFunc interface {
	Config() ScalarFuncConfig
	Executor() ScalarFuncExecutor
}
```

taniabogatsch:
Duckdb internally parallelizes scalar function execution, and each chunk is independently executed (on different threads, if available). So, indeed, we do not need states here.

I went over your feedback and pushed some changes. 
- [ ] Clean-up pass.
- [ ] Documentation of exported structs.
- [ ] Scalar function example(s).

Could you give another review (after these remaining steps)?

scalar_udf.go
```diff
@@ -0,0 +1,247 @@
...
+		return
+	}
+
+	// Execute the user-defined scalar function for each row.
+	executor := function.Executor()
+	values := make([]driver.Value, len(inputChunk.columns))
+	columnCount := len(values)
+
+	var err error
+	for rowIdx := 0; rowIdx < inputChunk.GetSize(); rowIdx++ {
```
JAicewizard, line 76: I think this currently calls cgo, if you put `inputChunk.GetSize()` outside of the loop it should improve performance

scalar_udf.go
```diff
@@ -0,0 +1,247 @@
...
+	"unsafe"
+)
+
+type rowFn func(args []driver.Value) (any, error)
+
+type ScalarFuncConfig struct {
+	InputTypeInfos []TypeInfo
+	ResultTypeInfo TypeInfo
+
+	VariadicTypeInfo    *TypeInfo
```
JAicewizard, line 28: Why is this a pointer? TypeInfo is already an interface which can be `nil`

scalar_udf.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func registerInputParams(config ScalarFuncConfig, f C.duckdb_scalar_function) error {
+	// Set variadic input parameters.
+	if config.VariadicTypeInfo != nil {
+		t := (*config.VariadicTypeInfo).logicalType()
+		C.duckdb_scalar_function_set_varargs(f, t)
+		C.duckdb_destroy_logical_type(&t)
+		return nil
+	}
```
JAicewizard, line 113: Would it be possible to register a function with both variadic args and non-variadic args?
If so, I don't think we shouldn't block users from registering with both variadic and regular args

scalar_udf.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+func registerResultParams(config ScalarFuncConfig, f C.duckdb_scalar_function) error {
+	if config.ResultTypeInfo == nil {
+		return errScalarUDFResultTypeIsNil
+	}
+	if config.ResultTypeInfo.InternalType() == TYPE_ANY {
+		return errScalarUDFResultTypeIsANY
+	}
```
JAicewizard, line 140: Is this because we do not support the ANY type? In that case I think the message should mention that we don't support it, since it seams reasonable to support this in the future

scalar_udf.go
JAicewizard: Sure this seems fine, I think I prefer how it is implemented for table functions but any way to add this functionality without a breaking change is good.

scalar_udf.go
```diff
@@ -0,0 +1,247 @@
...
+		setFuncError(function_info, getError(errAPI, err).Error())
+	}
+}
+
+//export scalar_udf_delete_callback
+func scalar_udf_delete_callback(extraInfo unsafe.Pointer) {
+	h := (*cgo.Handle)(extraInfo)
+	h.Delete()
+}
+
```
JAicewizard, line 105: Just putting this here, but has a bigger impact.
See https://github.com/golang/go/issues/68044 for the issue, but TLDR the go GC might deallocate whatever the handle is pointing to, so you need to pin it to make sure whatever it points to stays alive

scalar_udf.go
taniabogatsch: Oh, great catch!

scalar_udf.go
taniabogatsch: Yes, I'll add your `pinner` changes to the PR. I forgot about this; thanks for pointing it out.

scalar_udf.go
JAicewizard: Ah, that is really odd to not support the any type, but I guess if duckdb doesnt support it we won't have to either :)

taniabogatsch: In DuckDB internally, there are functions with the `ANY` return type. However, we must look at the input parameters during binding and resolve the `ANY` return type (in a custom bind function/callback). This callback is currently missing in the API on the duckdb side. Maybe in the future, if we add support in DuckDB, we can extend the `ScalarFuncExecutor` with the corresponding `ResolveReturnType` function type.

In `scalar_function.hpp`: 
```cpp	
//! The bind function (if any)
bind_scalar_function_t bind;
```

taniabogatsch:
Alright, I've implemented the review feedback (thanks!), and from my side, this should be ready to go in.
What do you think @JAicewizard?

PR #227, opened 2024-06-05
Add support for named parameters

ajzo90:
This PR adds support for named parameters. 
It tries to keeps old behaviour intact, but introduces some vague interpretation since it can mix named and non-named parameters. Possibly a better alternative is to check whether named or positional parameters are used and do strict validation. 

```
stmt, _ := db.Prepare("SELECT $foo, $bar")
_ = stmt.QueryRow(sql.Named("foo", 1), sql.Named("bar", "baz"))`
```

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -46,14 +46,37 @@ func (s *stmt) NumInput() int {
 }
 
 func (s *stmt) bind(args []driver.NamedValue) error {
-	if s.NumInput() != len(args) {
+	if s.NumInput() > len(args) {
 		return fmt.Errorf("incorrect argument count for command: have %d want %d", len(args), s.NumInput())
 	}
 
 	// FIXME (feature): we can't pass nested types as parameters (bind_value) yet
 
-	for i, v := range args {
-		switch v := v.Value.(type) {
+	// relaxed length check allow for unused parameters.
+	for i := 0; i < s.NumInput(); i++ {
+		name := C.duckdb_parameter_name(*s.stmt, C.idx_t(i+1))
+		paramName := C.GoString(name)
+		C.duckdb_free(unsafe.Pointer(name))
+		//t := C.duckdb_param_type(*s.stmt, C.idx_t(i+1))
+
+		// fallback on index position
+		var arg = args[i]
+
+		// override with ordinal if set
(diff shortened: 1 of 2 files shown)


statement.go
```diff
@@ -46,14 +46,37 @@ func (s *stmt) NumInput() int {
...
+				arg = v
+			}
+		}
+
+		// override with name if set
+		for _, v := range args {
+			if v.Name == paramName {
+				arg = v
+			}
+		}
```
marcboeker, line 77: Can we merge these two loops?

statement.go
```diff
@@ -46,14 +46,37 @@ func (s *stmt) NumInput() int {
...
 	// FIXME (feature): we can't pass nested types as parameters (bind_value) yet
 
-	for i, v := range args {
-		switch v := v.Value.(type) {
+	// relaxed length check allow for unused parameters.
+	for i := 0; i < s.NumInput(); i++ {
+		name := C.duckdb_parameter_name(*s.stmt, C.idx_t(i+1))
+		paramName := C.GoString(name)
+		C.duckdb_free(unsafe.Pointer(name))
+		//t := C.duckdb_param_type(*s.stmt, C.idx_t(i+1))
```
marcboeker, line 60: Please remove.

statement_test.go
```diff
@@ -20,6 +22,23 @@ func TestPrepareQuery(t *testing.T) {
 	defer rows.Close()
 }
 
+func TestPrepareQueryNamed(t *testing.T) {
+	db := openDB(t)
+	defer db.Close()
+	createFooTable(db, t)
+
+	stmt, err := db.PrepareContext(context.Background(), "SELECT $foo,$bar,$baz,$foo")
```
marcboeker, line 30: Please add spaces between args.

marcboeker (commented):
@ajzo90 Thanks for this PR. Please see my comments.

statement.go
ajzo90: It is in 3 passes to set different priorities. For example in case the client code or a wrapper library is setting/generating `NamedValues` with both a `Name` and an `Ordinal` as a fallback mechanism.

PR #233, opened 2024-06-07
Initial DataChunk API support

taniabogatsch:
This PR introduces a new `DataChunk` API. See the discussion [here](https://github.com/marcboeker/go-duckdb/pull/219). Ideally, all interaction between go-duckdb and duckdb's data chunks eventually goes through this API without compromising performance. As an outlook, in #219, I further used this API for go-duckdb's scanner.

As a first step, this PR moves all interaction between the `Appender` and data chunks to the `DataChunk` API and its setter functions, further decreasing the complexity of the `Appender` code.

This PR adds the following functions:
- `func (chunk *DataChunk) InitFromTypes(ptr unsafe.Pointer, types []C.duckdb_logical_type) error`
- `func (chunk *DataChunk) Destroy()`
- `func (chunk *DataChunk) SetSize() error`
- `func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error`

Additionally, I reverted the column index in the data chunk when returning an unsupported type error message to 0-based indexing.

### Questions
- I introduced a new `helper.go` file. However, I recommend renaming this file or moving its functions to a different place. What do you think?
- The `DataChunk`, and by implication, the `vector` structs are becoming universal components without ties to the `Appender`. Thus, I renamed `appender_vector.go` to `vector.go`. Also, I added a new file, `vector_setters.go`, to create some functionality separation and decrease file sizes. Is this a good approach?

### DIFF
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -1,0 +1,104 @@
+package duckdb
+
+/*
+#include <stdlib.h>
+#include <duckdb.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+// DataChunk storage of a DuckDB table.
+type DataChunk struct {
+	// data holds the underlying duckdb data chunk.
+	data C.duckdb_data_chunk
+	// columns is a helper slice providing direct access to all columns.
+	columns []vector
+	// columnNames holds the column names, if known.
+	columnNames []string
+}
+
+// InitFromTypes initializes a data chunk by providing its column types.
+func (chunk *DataChunk) InitFromTypes(ptr unsafe.Pointer, types []C.duckdb_logical_type) error {
+	columnCount := len(types)
+
+	// Initialize the callback functions to read and write values.
+	chunk.columns = make([]vector, columnCount)
+	var err error
+	for i := 0; i < columnCount; i++ {
+		if err = chunk.columns[i].init(types[i], i); err != nil {
+			break
+		}
+	}
+	if err != nil {
+		return err
+	}
+
+	logicalTypesPtr := (*C.duckdb_logical_type)(ptr)
+	chunk.data = C.duckdb_create_data_chunk(logicalTypesPtr, C.idx_t(columnCount))
+	C.duckdb_data_chunk_set_size(chunk.data, C.duckdb_vector_size())
+
+	// Initialize the vectors and their child vectors.
+	for i := 0; i < columnCount; i++ {
+		duckdbVector := C.duckdb_data_chunk_get_vector(chunk.data, C.idx_t(i))
+		chunk.columns[i].duckdbVector = duckdbVector
+		chunk.columns[i].getChildVectors(duckdbVector)
+	}
+	return nil
+}
+
+// Destroy the memory of a data chunk. This is crucial to avoid leaks.
+func (chunk *DataChunk) Destroy() {
+	C.duckdb_destroy_data_chunk(&chunk.data)
+}
+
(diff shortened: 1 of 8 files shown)


ajzo90:
Nice! A few suggestions on the public interface.

Perhaps `InitFromTypes` and `SetSize` should be private methods (for now)? `InitFromTypes` because it's using internal arguments and `SetSize` since it's likely not necessary to expose publicly?

`Destroy` can also be private until we have cases where it needs to be public. If/when it goes public I think the signature should be `Close() error` to follow go naming conventions.

taniabogatsch:
@ajzo90, good call, I've made them private. Do you have any other suggestions?

JAicewizard:
Small question, is there a reason projection support wasn't added? If not could this still be added via another PR? and if yet should we make a new type to support projection?

JAicewizard:
Also for table functions `SetSize` is must be exposed, as a tablefunction might not want to fill the entire chunk, I will expose this in my PR

vector.go
```diff
@@ -252,163 +267,52 @@ func (vec *vector) getChildVectors(vector C.duckdb_vector) {
...
-		childVector := vec.childVectors[i]
-		childName := vec.childNames[i]
-		childVector.fn(&childVector, rowIdx, m[childName])
-	}
-}
-
 func initPrimitive[T any](vec *vector, duckdbType C.duckdb_type) {
-	vec.fn = func(vec *vector, rowIdx C.idx_t, val any) {
+	vec.setFn = func(vec *vector, rowIdx C.idx_t, val any) {
+		vec.size++
```
JAicewizard, line 272: This seems wrong, if we start with an empty vector, and set row `50` to x, the size should be 50 (right?). If we do it like this the size would be 1, and we would have one row with an unset value at 1.

data_chunk.go
```diff
@@ -0,0 +1,93 @@
...
+	allEqual := true
+	maxSize := C.idx_t(chunk.columns[0].size)
+	for i := 0; i < len(chunk.columns); i++ {
+		if chunk.columns[i].size != maxSize {
+			allEqual = false
+		}
+		if chunk.columns[i].size > maxSize {
+			maxSize = chunk.columns[i].size
+		}
+	}
```
JAicewizard, line 86: This seems wrong, if the user wants to keep one column completely empty (all nulls), the size of this column would be 0 (unless they explicitly set everything to null). This would return an error, even though this seems perfectly reasonable.

JAicewizard (commented):
Sorry I am just reviewing this now, I wasn't aware of the PR and didn't have any time to review this week. In the case these need to be fixed, maybe a fix PR would be appropriate?

vector.go
taniabogatsch: Yes, that's a fair argument. I'll create a fix PR. Sorry for not tagging you on this PR! However, we always need to set the maximum. Now that I think about this, we probably want to decouple this from the vector entirely and change `func (chunk *DataChunk) setSize(size int)` to take an argument.

data_chunk.go
taniabogatsch: See my comment above.

vector.go
JAicewizard: I don't seem to have gotten a mail about this, must have gotten lost in the sea of spam github sends about failed CI. But thanks for the response! I'll rebase my code sometime after next week

taniabogatsch: No worries - :)

PR #237, opened 2024-06-20
Appender support for TIME, INTERVAL, HUGEINT, DECIMAL, ENUM, and MAP

taniabogatsch:
This PR introduces support for `TIME`, `INTERVAL`, `HUGEINT`, `DECIMAL`, `ENUM`, and `MAP` in the `Appender`. go-duckdb now supports all scannable types in the `Appender`.

### Other relevant changes

- I moved most primitive appender tests into `testTypes` to populate the table. That way, we're testing both appending to the table and scanning from it.
- For faster `SetValue` calls, we store additional information for `STRUCT`, `ENUM`, and `DECIMAL` in `vector`, if applicable.

### EDIT
I've extended this PR with some code paths to skip typecasting. This does not include more than one-level nested `LIST`s and `STRUCT`s, for which we still transform the input.

### EDIT EDIT
I'll move these changes into a different (follow-up) PR.

### Benchmark results

`BenchmarkTypes-10    	       4	 329119177 ns/op`

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -74,7 +74,7 @@ func NewAppenderFromConn(driverConn driver.Conn, schema, table string) (*Appende
 
 		// Ensure that we only create an appender for supported column types.
 		duckdbType := C.duckdb_get_type_id(a.types[i])
-		name, found := unsupportedAppenderTypeMap[duckdbType]
+		name, found := unsupportedTypeMap[duckdbType]
 		if found {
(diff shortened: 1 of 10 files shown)


taniabogatsch:
@JAicewizard, I've added some changes to the type casting for `SetValue[Unsafe]`. Is there a way to further speed this up?

PR #240, opened 2024-06-26
Move row scanning int the data chunk interface

taniabogatsch:
This PR expands the existing `DataChunk` interface with `GetValue`. I've moved all the scanning functionality out of `rows.go` and into the respective getter functions. The getter function matches the `fnGetVectorValue` type: `type fnGetVectorValue func(vec *vector, rowIdx C.idx_t) any.` During data chunk initialization, we now initialize both the setter and getter functions.

A possible optimization for multiple (same result) data chunks is to reuse the existing initialized vectors and change the duckdb vector and data. Even without that optimization, this PR should speed up scanning. I've run the `BenchmarkTypes` benchmark both on `main` and this PR (locally).

| Benchmark | main      | PR |
| ----------- | ----------- | ----------- |
| BenchmarkTypes-10 | 341,039,886 ns/op      | 271,966,542 ns/op       |

### DIFF
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -16,11 +16,18 @@ type DataChunk struct {
 	data C.duckdb_data_chunk
 	// columns is a helper slice providing direct access to all columns.
 	columns []vector
+	// columnNames holds the column names, if known.
+	columnNames []string
 }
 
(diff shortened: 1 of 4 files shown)


marcboeker:
@taniabogatsch Thanks for the improvement.

PR #242, opened 2024-07-02
add test cases to error tests

apocelipes:
Added some test cases selected from duckdb's test set. And added a test for the method `(*duckdb.Error).Is`. 

Replace "DuckDBError" in the help information with "(*duckdb.Error)".


marcboeker:
Thanks @apocelipes!

PR #247, opened 2024-07-11
docs: Support godoc examples for 'simple' and 'advanced' configurations, add GoDoc badge

dxh9845:
This PR adds Godoc examples on how to use the Goduckdb driver. These examples will be rendered inline on the DuckDB GoDoc page. 

Here's how the examples look when rendered locally: 
image

I've also added the `GoDoc` badge that will link directly to the documentation.

Please let me know if this is ok!


marcboeker:
@dxh9845 Thanks for your contribution. More examples is always a good idea. Do you think we can move them to the `examples/` folder and convert the existing examples also to the specific Godoc format to have a common place and format for all examples?

taniabogatsch:
We can link to the [godoc `examples` package ](https://pkg.go.dev/github.com/marcboeker/go-duckdb/examples) in the [DuckDB docs](https://duckdb.org/docs/api/go.html), and in the `README.md`. Then, we can add many different examples to the folder, and don't have to worry about cluttering the source folder with files.

Would that work @dxh9845?

PR #248, opened 2024-07-13
Add support for windows 

JAicewizard:
Pretty self-explanatory, this adds support for windows.

There are a couple issues, notably that duckdb doesn't build many extensions for mingw on windows. For this reason, I decided to statically link the json extension, as it feels essential. This leads to some platform differences, as explicitly installing/loading the json extension will result in an error.


taniabogatsch:
Amazing, thanks for working on this! I've moved some of your changes in duckdb to a new PR here to (hopefully) include them in the upcoming release: https://github.com/duckdb/duckdb/pull/13744
That way, we should be able to remove the patch from this PR. I'll review the other changes of this PR after merging the duckdb-side.

Makefile
```diff
@@ -28,6 +30,7 @@ DUCKDB_COMMON_BUILD_FLAGS := BUILD_SHELL=0 BUILD_UNITTESTS=0 DUCKDB_PLATFORM=any
 .PHONY: deps.darwin.amd64
 deps.darwin.amd64: duckdb
 	if [ "$(shell uname -s | tr '[:upper:]' '[:lower:]')" != "darwin" ]; then echo "Error: must run build on darwin"; false; fi
+	mkdir -p deps/darwin_amd64
```
taniabogatsch, line 33: Why is this `mkdir -p deps/respective_build` necessary in all the builds?

Makefile
```diff
@@ -52,14 +57,36 @@ deps.linux.amd64: duckdb
...
 	cp duckdb/build/release/libduckdb_bundle.a deps/linux_arm64/libduckdb.a
 
+.PHONY: deps.freebsd.amd64
+deps.windows.amd64: duckdb
+	if [ "$(shell uname -s | tr '[:upper:]' '[:lower:]')" != "mingw64_nt-10.0-20348" ]; then echo "Error: must run build on windows"; false; fi
+	mkdir -p deps/windows_amd64
+	
+	# this is just code copied from duckdb and fixed for windows. Would like to not change this, and use `make bundle-library` once its fixed.
+	cd duckdb && \
+	${DUCKDB_COMMON_BUILD_FLAGS} GENERATOR="-G \"MinGW Makefiles\"" BUILD_JSON=1 gmake release -j 2
```
taniabogatsch, line 73: I am undecided about compiling `JSON` into the build, as it diverges from the other builds... cc @marcboeker

Makefile
```diff
@@ -52,14 +57,36 @@ deps.linux.amd64: duckdb
...
 	cd duckdb && \
 	CC="aarch64-linux-gnu-gcc" CXX="aarch64-linux-gnu-g++" CFLAGS="-O3" CXXFLAGS="-O3" ${DUCKDB_COMMON_BUILD_FLAGS} make bundle-library -j 2
 	cp duckdb/build/release/libduckdb_bundle.a deps/linux_arm64/libduckdb.a
 
+.PHONY: deps.freebsd.amd64
+deps.windows.amd64: duckdb
+	if [ "$(shell uname -s | tr '[:upper:]' '[:lower:]')" != "mingw64_nt-10.0-20348" ]; then echo "Error: must run build on windows"; false; fi
+	mkdir -p deps/windows_amd64
+	
+	# this is just code copied from duckdb and fixed for windows. Would like to not change this, and use `make bundle-library` once its fixed.
```
taniabogatsch, line 71: I agree. But from my side, this is fine for this initial PR. We can turn it into a `FIXME: use bundle-library, once fixed for Windows` note.

cgo_static.go
```diff
@@ -1,13 +1,15 @@
-//go:build !duckdb_use_lib && (darwin || (linux && (amd64 || arm64)) || (freebsd && amd64))
+//go:build !duckdb_use_lib && (darwin || (linux && (amd64 || arm64)) || (freebsd && amd64) || (windows && amd64))
 
 package duckdb
 
 /*
+#cgo CPPFLAGS: -DDUCKDB_STATIC
```
taniabogatsch, line 6: Could you update once the respective duckdb PR is merged? Then, we can also remove the patch.

cgo_static.go
```diff
@@ -1,13 +1,15 @@
...
 #cgo LDFLAGS: -lduckdb
 #cgo darwin,amd64 LDFLAGS: -lc++ -L${SRCDIR}/deps/darwin_amd64
 #cgo darwin,arm64 LDFLAGS: -lc++ -L${SRCDIR}/deps/darwin_arm64
 #cgo linux,amd64 LDFLAGS: -lstdc++ -lm -ldl -L${SRCDIR}/deps/linux_amd64
 #cgo linux,arm64 LDFLAGS: -lstdc++ -lm -ldl -L${SRCDIR}/deps/linux_arm64
+#cgo windows,amd64 LDFLAGS: -lws2_32 -lwsock32 -lRstrtmgr -lstdc++ -lm --static -L${SRCDIR}/deps/windows_amd64
```
taniabogatsch, line 12: Out of curiosity - how did you know which flags to add?
I found this. Did you add them by looking at the error messages about missing libraries?
> For static linking, one has to specify all the libraries explicitly, including those required indirectly (i.e. by libraries that you use).

Makefile
JAicewizard: I don't know, I think it was mostly for consistency? It has been a while, I can remove it if you want

Makefile
JAicewizard: The problem is that (At the time) there were no json extension builds for this bulid. (I even think there were no extensions available at all). So compiling it with duckdb was the only way to make json available to the users, and I found json to be essential enough to include it. Maybe we should try again when 1.1 has released, hopefully also including the header changes?

cgo_static.go
JAicewizard: >  I found this. Did you add them by looking at the error messages about missing libraries?

Yes, exactly. I just looked at the error messages. Waiting half an hour every time wasn't ideal, but it worked :)

Makefile
JAicewizard: Oh I needed it because there didn't exist a directory yet. The usecase may be small, but I think in general it is nicer to not assume a directory already exists, even if we know it does. I can remove it if you want to,

taniabogatsch: That's a fair point, let's keep it 👍

Makefile
taniabogatsch: Yes, that's a good call - we can pick this up after `1.1.` and discuss the extensions.

cgo_static.go
taniabogatsch: Haha yes I can imagine 😄

yusufozturk:
@taniabogatsch any updates for this? We are waiting for this PR :) Thanks.

taniabogatsch:
Thanks for the ping - I'll merge the most recent changes and rerun everything. 🤞

taniabogatsch (approved):

cgo_static.go
albertobruin: Hi @JAicewizard, thanks for your contribution btw . I'm trying to build for windows with [goreleaser](https://github.com/goreleaser/goreleaser-cross)  and it's failing with 

```
/usr/bin/x86_64-w64-mingw32-ld: cannot find -lRstrtmgr
````

Do you know how can one get the libraries needed to compile?

JAicewizard: If your ditro doesn't ship this library you're out of luck for cross-compiling.
But you could try https://github.com/marcboeker/go-duckdb/pull/293 which made everything lowercase

albertobruin: Thanks a lot for your reply. Making the R lowercase took me to the next issue, which was adding `-lpthread`. now im fighting some std missing functions. Thanks a lot for your reply, really appreciate it.

PR #260, opened 2024-08-12
GitHub workflow actions: automated version check

deining:
GitHub workflow actions got outdated in the past (#259) .
In order to prevent this from happening again, this PR adds automated version check for GitHub workflow actions of this repo.
I hope you like it. Automated check can be enabled for Go dependencies, too (if wanted).

PR #261, opened 2024-08-13
Enable automated version check for go modules

deining:
This PR adds automated version check for Go modules, as proposed and confirmed [here](https://github.com/marcboeker/go-duckdb/pull/260).


marcboeker:
@deining Thanks!

PR #270, opened 2024-08-30
[Appender] duckdb_appender_flush before close for a meaningful error message

taniabogatsch:


marcboeker:
@taniabogatsch Welcome back and thanks!

PR #272, opened 2024-09-10
[Feature] Type interface

taniabogatsch:
This PR adds methods to expose `Type` creation. This is necessary for the table UDF (#201) and scalar UDF (#222) PRs.
There are also a few other changes here and there, but the gist is introducing the exported `Type`.

~cc @JAicewizard, @marcboeker, or others. I currently use the functions below. There is a switch for the primitive `reflect` types, some very basic functions for the non-Go types, and more complex functions for some types. However, there are a few things that I do not like about this approach.~
~- Usage of `reflect` as input argument.~
~- A lot of functions that could be a switch or a map lookup.~


JAicewizard:
I don't have much time ATM, but this weekend I will look at this and give a review.

type.go
taniabogatsch, line 1: I've reworked the initial approach. We now export `Type`, which wraps the internal duckdb type. With that, we can export each internal type.

Then, `TypeInfo` holds any information necessary for type creation (in UDFs).

type_info.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if childInfo.t == TYPE_INVALID {
+		return t, getError(errAPI, errInvalidChildType)
+	}
+	t.childTypes[0] = childInfo
+	return t, nil
+}
+
+// StructTypeInfo returns STRUCT type information.
+func StructTypeInfo(typeInfos []TypeInfo, names []string) (TypeInfo, error) {
```
JAicewizard, line 101: Whats the benefit of this over `[]StructField`? I think its a better API enforcing equal type infos and names. But its very minor, just a nitpick

type_info.go
```diff
@@ -0,0 +1,244 @@
...
+	case TYPE_ENUM:
+		return typeInfo.logicalEnumType(), nil
+	case TYPE_LIST:
+		return typeInfo.logicalListType(), nil
+	case TYPE_STRUCT:
+		return typeInfo.logicalStructType(), nil
+	case TYPE_MAP:
+		return typeInfo.logicalMapType(), nil
+	}
+	return nil, unsupportedTypeError(unknownTypeErrMsg)
```
JAicewizard, line 173: If we make TypeInfo an interface with `logicalType` as method, this would be handled by the type system. I think it would then also be reasonably to say the `primitiveType` type could then assume the type is valid and supported, and the error could be removed. I think that makes sense, since any issues should have been caught when creating the `TypeInfo` using the methods above, once that is done there should not be any errors.

vector.go
```diff
@@ -26,88 +25,71 @@ type vector struct {
...
 	childVectors []vector
 
-	// The child names of STRUCT vectors.
-	childNames []string
-	// The dictionary for ENUM types.
-	dict map[string]uint32
-	// The  width of DECIMAL types.
-	width uint8
-	// The scale of DECIMAL types.
-	scale uint8
```
JAicewizard, line 41: Nice, these always seems unnecessary to me :)

JAicewizard (commented):
These are my comments for now, I will try to rebase this weekend, and find out whether or not I find any issues with this PR, but I doubt it.

type_info.go
taniabogatsch: That's a good point; I'll add it. 👍

type_info.go
taniabogatsch: Great suggestion, I've turned `TypeInfo` into an interface.

vector.go
taniabogatsch: Yes, I agree. Sadly, we do need some of them in the setters and getters. But at least they are in their dedicated `struct` now.

type_info.go
```diff
@@ -0,0 +1,240 @@
...
+type typeInfo struct {
+	baseTypeInfo
+	childTypes []TypeInfo
+	enumNames  []string
+}
+
+// TypeInfo is an interface for DuckDB types.
+type TypeInfo interface {
+	logicalType() C.duckdb_logical_type
+}
```
taniabogatsch, line 70: I've turned `TypeInfo` into an `interface`. My only concern is that developers might run into `nil` pointer dereferences when passing uninitialized interfaces to functions if we no longer return an error instead. Is this fine for idiomatic Go? Or should each function that accepts an interface also contain a `nil` check and return an error?

The same question applies to `type StructEntry interface`.

JAicewizard: Hmm, you raise a good point. I think the best way it indeed to either return an error, or default to `any` (which is just a type without any typeinfo right?).

 for `type StructEntry interface` I think it should just be a `struct{info Typeinfo, name string}` avoiding this whole issue (with a nil typeinfo still being `any`).
 
 Sorry if I'm a bit opinionated, but I think this is a very nice API :)

taniabogatsch: No worries; I appreciate the discussion and input.
I'll then add `nil` checks and return an appropriate error.

W.r.t. `StructEntry`, I prefer using the interface, as we stay flexible with the internals and can add error checks during creation, like the empty name check.

taniabogatsch:
Let me know how using the interface for the table UDFs goes. If it works well, then I think this is ready to merge.

JAicewizard:
I have one small nit, it would be really nice to have an error-less way to create supported types. That is the reason I implemented ways to create duckdb types using generics, since it allows statically checking that the type will result in a supported duckdb type.

One example is the following piece of code:
```
func (d *incTableUDF) GetFunction() RowTableFunction {
	return RowTableFunction{
		Config: TableFunctionConfig{
			Arguments: []Type{NewDuckdbType[int64]()},
		},
		BindArguments: BindIncTableUDF,
	}
}
```
which is now the following:
```
func (d *incTableUDF) GetFunction() RowTableFunction, error {
	t, err := NewTypeInfo(TYPE_BIGINT)
	if err != nil {
		return RowTableFunction{}, err
	}
	return RowTableFunction{
		Config: TableFunctionConfig{
			Arguments: []Type{},
		},
		BindArguments: BindIncTableUDF,
	}, nil
}
```
+ extra error handling in go-duckdb

which is significantly longer. Since table functions already contain quite a bit of boilerplate (on the users end) I would prefer to have some error-less way to create basic types. This could also remove the error return parameter on the `Config` method for the scalar functions, but I think there is much less boilerplate so it is less of a problem.

JAicewizard:
Additionally, using generics to create a new type using generics/reflection allows easily creating deeply nested types since it can be infered from the go type

taniabogatsch:
> I have one small nit, it would be really nice to have an error-less way to create supported types. That is the reason I implemented ways to create duckdb types using generics, since it allows statically checking that the type will result in a supported duckdb type.

I understand the reasoning behind this. When I tried to move your code into this PR, though, I encountered a few issues with using generics.

For any timestamp-based duckdb type, we cannot infer the intended type from the Go type (to the best of my knowledge).
```Go
case time.Time:
	// TYPE_TIMESTAMP, TYPE_TIMESTAMP_S, TYPE_TIMESTAMP_MS, TYPE_TIMESTAMP_NS, TYPE_TIMESTAMP_TZ,
     // TYPE_DATE, TYPE_TIME?
```

Also, for `UUID` and `byte`, I don't know how to best determine whether it is a `UUID` or a `BLOB`.

taniabogatsch:
Maybe a way to use generics is to add the timestamp types to go-duckdb. We already do this for `Interval`, `Decimal`, `UUID`, etc. 🤔

taniabogatsch:
My preference would be errorless code as well. But after thinking more about it, I am unsure how to solve any ambiguity issues, both with the currently supported types and any potential future types.

For example, how would we infer an `ENUM` type? Or, if we were to add support for `ARRAY`, how would we distinguish between a slice with three (fixed-size array) elements, or a `LIST`.

So, while the current approach introduces some error handling and more code for nested type creation, it is also much more flexible and robust... 🤔 

An alternative to deciding between one of the two approaches might be to add a fast path for primitive types or an `UnsafeNewXX` function that omits errors. But that would add more complexity to the API for users and us to maintain.

taniabogatsch:
Here is a snippet for using generics for type detection.
https://github.com/marcboeker/go-duckdb/pull/201#discussion_r1760934637

cc @marcboeker

PR #274, opened 2024-09-10
Add profiling support

taniabogatsch:
Depends on https://github.com/duckdb/duckdb/pull/13844.
I'll move this from `Draft` to a normal pull request after [DuckDB's bug fix release](https://duckdb.org/docs/dev/release_calendar.html#upcoming-releases).

### DIFF
--- a/profiling.go
+++ b/profiling.go
@@ -4,6 +4,7 @@ package duckdb
 #include <duckdb.h>
 */
 import "C"
+
 import (
 	"unsafe"
 )
--- a/profiling.go
+++ b/profiling.go
@@ -1,0 +1,67 @@
+package duckdb
+
+/*
+#include <duckdb.h>
+*/
+import "C"
+import (
+	"unsafe"
+)
+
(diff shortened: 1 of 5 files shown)


profiling.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"unsafe"
+)
+
+type ProfilingInfo struct {
+	Metrics  map[string]string
+	Children []ProfilingInfo
+}
+
+func GetProfilingInfo(driverConn any) (ProfilingInfo, error) {
```
marcboeker, line 17: Would it make sense to add a comment to explain what profiling information is returned (e.g. link to the example at https://duckdb.org/docs/dev/profiling.html#setting-custom-metrics-examples) or to add a section to the README?

marcboeker:
@taniabogatsch Great new feature, thanks!

profiling.go
taniabogatsch: Yes, good point 👍

PR #276, opened 2024-09-18
Add generic setters to void casting values to any

JAicewizard:
Casting to any can be a significant overhead for table functions, as it usually requires an allocation.

In addition, the new functions also handle all possible castings that we support, so I think that we can also remove `tryCast` completely saving 200 linex, and I suspect we can even remove the `setFn` field, saving even more. Maybe someone can look at that to see if that sounds reasonable?

### DIFF
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -58,19 +58,8 @@ func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
 	}
 	column := &chunk.columns[colIdx]
 
-	// Ensure that the types match before attempting to set anything.
-	// This is done to prevent failures 'halfway through' writing column values,
-	// potentially corrupting data in that column.
-	// FIXME: Can we improve efficiency here? We are casting back-and-forth to any A LOT.
-	// FIXME: Maybe we can make columnar insertions unsafe, i.e., we always assume a correct type.
-	v, err := column.tryCast(val)
-	if err != nil {
-		return addIndexToError(err, colIdx)
-	}
-
 	// Set the value.
-	column.setFn(column, C.idx_t(rowIdx), v)
-	return nil
+	return column.setFn(column, C.idx_t(rowIdx), val)
 }
 
 func (chunk *DataChunk) initFromTypes(ptr unsafe.Pointer, types []C.duckdb_logical_type, writable bool) error {
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -62,8 +62,9 @@ func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
 	return column.setFn(column, C.idx_t(rowIdx), val)
 }
 
-// SetValue writes a single value to a column in a data chunk.
-// Note that this requires casting the type for each invocation.
+// SetChunkValue writes a single value to a column in a data chunk.
+// The difference with `chunk.SetValue` is that `SetChunkValue` does not
+// require that the value be cast to `any` (implicitly), and thus
 // NOTE: Custom ENUM types must be passed as string.
 func SetChunkValue[T any](chunk DataChunk, colIdx int, rowIdx int, val T) error {
 	if colIdx >= len(chunk.columns) {
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -50,7 +50,8 @@ func (chunk *DataChunk) GetValue(colIdx int, rowIdx int) (any, error) {
 	return column.getFn(column, C.idx_t(rowIdx)), nil
 }
 
-// SetValue writes a single value to a column in a data chunk. Note that this requires casting the type for each invocation.
+// SetValue writes a single value to a column in a data chunk.
+// Note that this requires casting the type for each invocation.
 // NOTE: Custom ENUM types must be passed as string.
 func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
 	if colIdx >= len(chunk.columns) {
@@ -64,7 +65,7 @@ func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
 
 // SetChunkValue writes a single value to a column in a data chunk.
 // The difference with `chunk.SetValue` is that `SetChunkValue` does not
-// require that the value be cast to `any` (implicitly), and thus
+// require casting the value to `any` (implicitly).
 // NOTE: Custom ENUM types must be passed as string.
 func SetChunkValue[T any](chunk DataChunk, colIdx int, rowIdx int, val T) error {
 	if colIdx >= len(chunk.columns) {
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -62,6 +62,16 @@ func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
 	return column.setFn(column, C.idx_t(rowIdx), val)
 }
 
+// SetValue writes a single value to a column in a data chunk.
+// Note that this requires casting the type for each invocation.
+// NOTE: Custom ENUM types must be passed as string.
+func SetChunkValue[T any](chunk DataChunk, colIdx int, rowIdx int, val T) error {
+	if colIdx >= len(chunk.columns) {
+		return getError(errAPI, columnCountError(colIdx, len(chunk.columns)))
+	}
+	return setVectorVal(&chunk.columns[colIdx], C.idx_t(rowIdx), val)
+}
+
 func (chunk *DataChunk) initFromTypes(ptr unsafe.Pointer, types []C.duckdb_logical_type, writable bool) error {
 	// NOTE: initFromTypes does not initialize the column names.
 	columnCount := len(types)
--- a/vector.go
+++ b/vector.go
@@ -239,27 +239,27 @@ func (vec *vector) init(logicalType C.duckdb_logical_type, colIdx int) error {
 
 	switch t {
 	case TYPE_BOOLEAN:
-		initPrimitive[bool](vec, t)
+		initBool(vec, TYPE_BOOLEAN)
 	case TYPE_TINYINT:
-		initPrimitive[int8](vec, t)
+		initNumeric[int8](vec, TYPE_TINYINT)
 	case TYPE_SMALLINT:
-		initPrimitive[int16](vec, t)
+		initNumeric[int16](vec, TYPE_SMALLINT)
 	case TYPE_INTEGER:
-		initPrimitive[int32](vec, t)
+		initNumeric[int32](vec, TYPE_INTEGER)
 	case TYPE_BIGINT:
-		initPrimitive[int64](vec, t)
+		initNumeric[int64](vec, TYPE_BIGINT)
 	case TYPE_UTINYINT:
-		initPrimitive[uint8](vec, t)
+		initNumeric[uint8](vec, TYPE_UTINYINT)
 	case TYPE_USMALLINT:
-		initPrimitive[uint16](vec, t)
+		initNumeric[uint16](vec, TYPE_USMALLINT)
 	case TYPE_UINTEGER:
-		initPrimitive[uint32](vec, t)
+		initNumeric[uint32](vec, TYPE_UINTEGER)
 	case TYPE_UBIGINT:
-		initPrimitive[uint64](vec, t)
+		initNumeric[uint64](vec, TYPE_UBIGINT)
 	case TYPE_FLOAT:
-		initPrimitive[float32](vec, t)
+		initNumeric[float32](vec, TYPE_FLOAT)
 	case TYPE_DOUBLE:
-		initPrimitive[float64](vec, t)
+		initNumeric[float64](vec, TYPE_DOUBLE)
 	case TYPE_TIMESTAMP, TYPE_TIMESTAMP_S, TYPE_TIMESTAMP_MS, TYPE_TIMESTAMP_NS, TYPE_TIMESTAMP_TZ:
 		vec.initTS(t)
 	case TYPE_DATE:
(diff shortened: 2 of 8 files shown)


JAicewizard:
Update: Removed all casting logic other than the new set interface

vector.go
```diff
@@ -424,12 +250,12 @@ func (vec *vector) initCString(t Type) {
...
 	}
-	vec.setFn = func(vec *vector, rowIdx C.idx_t, val any) {
+	vec.setFn = func(vec *vector, rowIdx C.idx_t, val any) error {
 		if val == nil {
 			vec.setNull(rowIdx)
-			return
+			return nil
 		}
-		vec.setCString(rowIdx, val)
+		return setString(vec, rowIdx, val)
```
taniabogatsch, line 258: I initially called this `initCString` and `setCString` to imply that this addresses both `VARCHAR` and `BLOB`.
It wasn't a good name, so maybe we can find a better one now. Otherwise, let's also rename `initCString` to `initString`.

vector_setters.go
```diff
@@ -32,108 +33,319 @@ func setPrimitive[T any](vec *vector, rowIdx C.idx_t, v T) {
...
+		}
+	case *big.Int:
+		if v == nil {
+			return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
+		}
+		if v.IsUint64() {
+			fv = T(v.Uint64())
+		} else {
+			fv = T(v.Int64())
+		}
```
taniabogatsch, line 78: Let's remove these cases, as we currently do not use them.

vector_setters.go
```diff
@@ -32,108 +33,319 @@ func setPrimitive[T any](vec *vector, rowIdx C.idx_t, v T) {
 	xs[rowIdx] = v
 }
 
-func (vec *vector) setTS(t Type, rowIdx C.idx_t, val any) {
-	v := val.(time.Time)
+// TODO: support Decimal here and bool and hugeint
```
taniabogatsch, line 36: Let's remove this comment and add a `FIXME` to the relevant `set...` functions. We can pick this up later.

vector_setters.go
```diff
@@ -32,108 +33,319 @@ func setPrimitive[T any](vec *vector, rowIdx C.idx_t, v T) {
...
+			fv = T(v.Value.Int64())
+		}
+	default:
+		return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
+	}
+	setPrimitive(vec, rowIdx, fv)
+	return nil
+}
+
+func setBool[S any](vec *vector, rowIdx C.idx_t, val S) error {
```
taniabogatsch, line 95: I don't understand why we need so many cases here. Shouldn't the following be sufficient?
```Go
func setBool[S any](vec *vector, rowIdx C.idx_t, val S) error {
	var fv bool
	switch v := any(val).(type) {
	case bool:
		fv = v
	default:
		return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
	}
	setPrimitive(vec, rowIdx, fv)
	return nil
}
```

vector_setters.go
```diff
@@ -32,108 +33,319 @@ func setPrimitive[T any](vec *vector, rowIdx C.idx_t, v T) {
...
+	interval2.micros = C.int64_t(interval.Micros)
+	setPrimitive(vec, rowIdx, interval2)
+	return nil
 }
 
-func (vec *vector) setHugeint(rowIdx C.idx_t, val any) {
-	v := val.(*big.Int)
-	hugeInt, _ := hugeIntFromNative(v)
-	setPrimitive(vec, rowIdx, hugeInt)
+func setHugeint[S any](vec *vector, rowIdx C.idx_t, val S) error {
```
taniabogatsch, line 213: Also here, why are we adding the extra cases?
And, since we return errors now, we do not need to omit the error in `fv, _ = hugeIntFromNative(v)` any longer.
```Go
func setHugeint[S any](vec *vector, rowIdx C.idx_t, val S) error {
	var fv C.duckdb_hugeint
	switch v := any(val).(type) {
	case *big.Int:
		if v == nil {
			return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
		}
		fv, _ = hugeIntFromNative(v)
	default:
		return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
	}
	setPrimitive(vec, rowIdx, fv)
	return nil
}
```

vector_setters.go
```diff
@@ -175,5 +429,75 @@ func (vec *vector) setMap(rowIdx C.idx_t, val any) {
...
+		return setEnum[S](vec, rowIdx, val)
+	case TYPE_LIST:
+		return setList[S](vec, rowIdx, val)
+	case TYPE_STRUCT:
+		return setStruct[S](vec, rowIdx, val)
+	case TYPE_UUID:
+		return setUUID[S](vec, rowIdx, val)
+	default:
+		return unsupportedTypeError("Could not convert type " + reflect.TypeOf(val).Name() + " to a duckdb type")
+	}
```
taniabogatsch, line 502: This function is currently unused, so we should remove it to avoid adding untested/dead code.

vector_setters.go
```diff
@@ -175,5 +429,75 @@ func (vec *vector) setMap(rowIdx C.idx_t, val any) {
 		i++
 	}
 
-	vec.setList(rowIdx, list)
+	setList(vec, rowIdx, list)
```
taniabogatsch, line 432: `setList` now returns an error.
```Go
err := setList(...)
return err
```

vector_setters.go
```diff
@@ -151,21 +363,63 @@ func (vec *vector) setList(rowIdx C.idx_t, val any) {
...
 
-func (vec *vector) setStruct(rowIdx C.idx_t, val any) {
-	m := val.(map[string]any)
+func setStruct[S any](vec *vector, rowIdx C.idx_t, val S) error {
+	var m map[string]any
+	switch v := any(val).(type) {
+	case map[string]any:
+		m = v
+	default:
+		//TODO type.Kind() == reflect.Map??
```
taniabogatsch, line 380: I don't think we should mix `STRUCT` and `MAP` here. Or maybe I misunderstand your `TODO`?

taniabogatsch (changes requested):
Hi @JAicewizard, thanks for the PR; this is great!
I've left some nits. 😄 

Did you happen to run the benchmark `func BenchmarkTypes(b *testing.B)` in `types_test.go`? I'd be curious about the improvements.

I also saw that you added code to auto-cast between some types, e.g., in `setBool`. I am unsure if we should add that or whether it should happen on the user side. Also because we currently only support it for some types... 🤔

vector.go
```diff
@@ -239,27 +48,27 @@ func (vec *vector) init(logicalType C.duckdb_logical_type, colIdx int) error {
...
+		initNumeric[uint32](vec, TYPE_UINTEGER)
 	case TYPE_UBIGINT:
-		initPrimitive[uint64](vec, t)
+		initNumeric[uint64](vec, TYPE_UBIGINT)
 	case TYPE_FLOAT:
-		initPrimitive[float32](vec, t)
+		initNumeric[float32](vec, TYPE_FLOAT)
 	case TYPE_DOUBLE:
-		initPrimitive[float64](vec, t)
+		initNumeric[float64](vec, TYPE_DOUBLE)
```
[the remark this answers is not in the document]
JAicewizard: oh huh Im not sure why that happened, wasn't intentional

vector.go
JAicewizard: `setBytes` and `initBytes`? Since that is the underlying datastructure for both that seems sensible

vector_setters.go
JAicewizard: The bool and bigint case? Sure, I mostly added all the cases because I thought it would be nice to support casting from as many types as possible, but I can reduce the amount of cases if you want. Should I just reduce the supported types to what was possible before?

vector_setters.go
JAicewizard: copy-paste from above: Sure, I mostly added all the cases because I thought it would be nice to support casting from as many types as possible, but I can reduce the amount of cases if you want. Should I just reduce the supported types to what was possible before?

vector_setters.go
JAicewizard: same as above

vector.go
taniabogatsch: Yes, perfect 👍

vector_setters.go
taniabogatsch: I think either 
1. We add code testing all these new cases or
2. We remove them for now and potentially add them later with respective tests.

My main issue is that our tests do not trigger these paths (afaik).

vector_setters.go
JAicewizard: Oh I just wanted to support more kinds of maps besides `map[string]any`, which I think will be nice. Since duckdb structs are not always known at compiletime, I think that allowing the user to use a types map would give the user more flexibility when working with duckdb structs. Only allowing go structs is very restrictive (since it has to be known at compile time), and users might want more type information than having to use a `map[string]any`

vector_setters.go
JAicewizard: Sure, I can also add a generic set methods to the chunk API? Since that is what it is intended for as well

vector_setters.go
taniabogatsch: > allowing the user to use a types map would give the user more flexibility when working with duckdb structs. Only allowing go structs is very restrictive

Ah, got you! 👍 I think for this PR we can turn it into a `FIXME` and potentially add this later. It does sound like a nice idea to give flexibility.

vector_setters.go
taniabogatsch: Either that (plus the tests) or you can add it with the table UDFs.

vector_setters.go
JAicewizard: Actually I think it would be nice to be able to set hugeints with normal ints, working with `big.Int` can be cumbersome and slow, if the user knows they won't hit the limit of normal ints it would be nice if we don't force them to use `big.Int`. But I will add test cases for this

vector_setters.go
JAicewizard: nope, I dont think these paths are tested, but the tests seem trivial. Ill just remove it for now, can always be added later.

vector_setters.go
taniabogatsch: Yes, I agree. Cool.

vector_setters.go
```diff
@@ -32,108 +33,264 @@ func setPrimitive[T any](vec *vector, rowIdx C.idx_t, v T) {
...
+	case *big.Int:
+		if v == nil {
+			return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
+		}
+		fv, _ = hugeIntFromNative(v)
+	case Decimal:
+		if v.Value == nil {
+			return castError(reflect.TypeOf(val).String(), reflect.TypeOf(fv).String())
+		}
+		fv, _ = hugeIntFromNative(v.Value)
```
taniabogatsch, line 200: Let's not omit the error returned by `hugeIntFromNative(big.NewInt(int64(v)))`.
```Go
if fv, err = hugeIntFromNative(big.NewInt(int64(v))); err != nil {
return err
}

File: vector_setters.go [hunk 11]
```diff
@@ -175,5 +374,74 @@ func (vec *vector) setMap(rowIdx C.idx_t, val any) {
...
+	case TYPE_ENUM:
+		return setEnum[S](vec, rowIdx, val)
+	case TYPE_LIST:
+		return setList[S](vec, rowIdx, val)
+	case TYPE_STRUCT:
+		return setStruct[S](vec, rowIdx, val)
+	case TYPE_UUID:
+		return setUUID[S](vec, rowIdx, val)
+	default:
+		return unsupportedTypeError("Could not convert type " + reflect.TypeOf(val).Name() + " to a duckdb type")
```
taniabogatsch, line 445: Let's use the same error message code as in `init`. Then, we can also remove the `TODO...` and `UHUGEINT` comments, as we already have them in `unsupportedTypeToStringMap`.
```Go
	name, inMap := unsupportedTypeToStringMap[t]
	if inMap {
		return addIndexToError(unsupportedTypeError(name), colIdx)
	}

// ... and
	default:
		return addIndexToError(unsupportedTypeError(unknownTypeErrMsg), colIdx)
```

vector_setters.go
```diff
@@ -175,5 +374,74 @@ func (vec *vector) setMap(rowIdx C.idx_t, val any) {
...
+	default:
+		return castError(reflect.TypeOf(val).String(), reflect.TypeOf(uuid).String())
+	}
+	hi := uuidToHugeInt(uuid)
+	setPrimitive(vec, rowIdx, hi)
+	return nil
+}
+
+// TODO: Support ARRAY UNION BIT ANY VARINT,
+func setVectorVal[S any](vec *vector, rowIdx C.idx_t, val S) error {
```
taniabogatsch, line 394: I don't see any tests for the switch here, but I realize that will be tricky without the respective data chunk. We can keep this, but let's later add a test for each type in the table UDFs. Similar to `func TestAllTypesScalarUDF(t *testing.T)`.

taniabogatsch (changes requested):
Just a few more comments and this should be ready to merge. :)

JAicewizard:
> Did you happen to run the benchmark func BenchmarkTypes(b *testing.B) in types_test.go? I'd be curious about the improvements.

I did this just now, but most of that benchmark is spend making queries and not actually setting the values using the appender. But I checked using pprof and the time spend setting the values went from 11% to 8%. Not sure how reliable that is. Most of the time is spend in duckdb, so we cannot really improve that.

vector_setters.go
JAicewizard: Yes, I agree that would be good, will resolve this for now

data_chunk.go
```diff
@@ -58,19 +58,18 @@ func (chunk *DataChunk) SetValue(colIdx int, rowIdx int, val any) error {
...
 	// Set the value.
-	column.setFn(column, C.idx_t(rowIdx), v)
-	return nil
+	return column.setFn(column, C.idx_t(rowIdx), val)
+}
+
+// SetValue writes a single value to a column in a data chunk.
+// Note that this requires casting the type for each invocation.
+// NOTE: Custom ENUM types must be passed as string.
+func SetChunkValue[T any](chunk DataChunk, colIdx int, rowIdx int, val T) error {
```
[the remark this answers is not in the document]
JAicewizard: I changed the documentation, but I don't think we should name it `TypedXxXx` because I don't think being typed, warrants being represented in the name. The current name is simple, concise, and descriptive of what the function does

vector_setters.go
JAicewizard: should I implement this? or do you disagree

data_chunk.go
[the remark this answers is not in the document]
taniabogatsch: Fair point 👍

vector_setters.go
JAicewizard: Ah sorry your previous message hadn't shown up on my screen yet, I thought github always live-updated things but that never really seems to work. Made it a fixme

taniabogatsch:
Alright, thanks a lot for your work on this! I'll merge it once the tests pass.

taniabogatsch (approved):

JAicewizard:
I wrote some benchmarks, and they showed a 20% in time spend appending. I also uncovered some crashes however, which should be investigated. I will open a PR with the benchmark, but I don't have a fix for the crashes yet

PR #278, opened 2024-09-20
Add benchmark and fix list resizing

JAicewizard:
When resizing, we need to update the vector data pointer


taniabogatsch (approved):
Great, thanks for fixing this! I've unified the new benchmark code and run the formatter. Once the tests pass, this should be ready to go in.

PR #294, opened 2024-10-28
Add support for renaming struct fields

asabil:
Add support for renaming struct fields by using the `db` struct tag.

### DIFF
--- a/appender_test.go
+++ b/appender_test.go
@@ -20,7 +20,7 @@ import (
 )
 
 type simpleStruct struct {
-	A int32
+	A int32 `db:"a"`
 	B string
 }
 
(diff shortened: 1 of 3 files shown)


appender_test.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 type simpleStruct struct {
-	A int32
+	A int32 `db:"a"`
 	B string
 }
```
taniabogatsch, line 25: Let's add another test for a `struct` like the one below. It should return a duplicate field error.

```Go
type myStruct struct {
	A int64 `db:"Duplicate"`
	Duplicate int64 
}
```

And for completeness, another one like this:
```Go
type myStruct struct {
	a         int64 `db:"does_not_exist"`
	B int64
}
```

taniabogatsch (changes requested):
Cool - I've added a small comment. Otherwise, this looks good!

appender_test.go
asabil: I added a handling for the duplicate keys case. However I am not sure how to handle the missing key case as it seems to be unhandled even without field renaming.

Any suggestions on how to proceed?

Thanks

taniabogatsch: Ah, sorry, I should've phrased my message better. However, it is not strictly necessary, since we already test that path later.

> And for completeness, another **error-test** like this:

taniabogatsch (approved):

taniabogatsch:
Thanks!

PR #295, opened 2024-10-29
Fix comparison against nil for the UUID type

asabil:


appender_test.go
taniabogatsch, line 1: Let's keep the previous case (`var r UUID` instead of `var r *UUID`) for completeness. 
We can modify the test like so:
```Go
func TestAppenderUUID(t *testing.T) {
	t.Parallel()
	c, con, a := prepareAppender(t, `CREATE TABLE test (id UUID)`)

	id := UUID(uuid.New())
	otherId := UUID(uuid.New())
	require.NoError(t, a.AppendRow(id))
	require.NoError(t, a.AppendRow(otherId))
	require.NoError(t, a.AppendRow((*UUID)(nil)))
	require.NoError(t, a.AppendRow(nil))
	require.NoError(t, a.Flush())

	// Verify results.
	res, err := sql.OpenDB(c).QueryContext(context.Background(), `SELECT id FROM test`)
	require.NoError(t, err)

	i := 0
	for res.Next() {
		if i == 0 {
			var r UUID
			require.NoError(t, res.Scan(&r))
			require.Equal(t, id, r)
		} else {
			var r *UUID
			require.NoError(t, res.Scan(&r))
			if i == 1 {
				require.Equal(t, otherId, *r)
			} else {
				require.Nil(t, r)
			}
		}
		i++
	}
	cleanupAppender(t, c, con, a)
}
```

taniabogatsch (changes requested):
Thanks! I only have a small remark w.r.t. the test.

appender_test.go
asabil: Thanks, updated the test with your suggestion

taniabogatsch (approved):

taniabogatsch:
Thanks!

PR #301, opened 2024-10-31
update arrow to v18

loicalleyne:
"github.com/apache/arrow/go/v17/arrow" => "github.com/apache/arrow-go/v18/arrow"


taniabogatsch:
Thanks!

PR #304, opened 2024-11-05
TIMETZ support and fixes around times

taniabogatsch:


taniabogatsch:
@marcboeker, should we merge this before tagging a new version? To include these fixes in the next released version?

PR #306, opened 2024-11-07
ARRAY type support

taniabogatsch:
This PR adds support for the `ARRAY` type.

**Appending `ARRAY` columns**

The `Appender` can append both slices and arrays to `ARRAY` columns. 
When passing a slice, the child count must match the array size.

**Scanning `ARRAY` columns**

Scanning into a slice directly or into a `Composite` array, e.g., `Composite[[3]int32]` is possible.


rows.go
```diff
@@ -127,6 +127,8 @@ func (r *rows) ColumnTypeScanType(index int) reflect.Type {
 		return reflect.TypeOf(map[string]any{})
 	case TYPE_MAP:
 		return reflect.TypeOf(Map{})
+	case TYPE_ARRAY:
+		return reflect.TypeOf([]any{})
```
taniabogatsch, line 131: Is there any better way to reflect the `ARRAY` type here?

marcboeker: Maybe sth. like:

```
reflect.TypeOf([0]any{})
```

Otherwise it would be a slice I think.

type_info.go
```diff
@@ -232,6 +235,25 @@ func NewMapInfo(keyInfo TypeInfo, valueInfo TypeInfo) (TypeInfo, error) {
...
 }
 
+// NewArrayInfo returns ARRAY type information.
+// childInfo contains the type information of the ARRAY's elements.
+// size is the ARRAY's fixed size.
+func NewArrayInfo(childInfo TypeInfo, size uint64) (TypeInfo, error) {
+	if childInfo == nil {
+		return nil, getError(errAPI, interfaceIsNilError("childInfo"))
+	}
+	if size == 0 {
```
marcboeker, line 245: Edge case, but I think there can also be an array of size 0 in Go, even if this makes no sense.

type_info_test.go
```diff
@@ -226,6 +251,10 @@ func TestErrTypeInfo(t *testing.T) {
 	nilStructEntry, err := NewStructEntry(nil, "hello")
 	require.NoError(t, err)
 
+	// Invalid ARRAY entry.
+	_, err = NewArrayInfo(validInfo, 0)
```
marcboeker, line 255: We do not allow arrays of zero length? I know, this would make no sense in the real world but should we block this even if it is valid (at least in Go - no idea how DuckDB behaves).

vector_setters.go
```diff
@@ -413,6 +385,57 @@ func setMap[S any](vec *vector, rowIdx C.idx_t, val S) error {
...
+		return err
+	}
+	if len(array) != int(vec.arrayLength) {
+		return invalidInputError(strconv.Itoa(len(array)), strconv.Itoa(int(vec.arrayLength)))
+	}
+	return setSliceChild(vec, array, rowIdx*C.idx_t(vec.arrayLength))
+}
+
+func extractSlice[S any](vec *vector, val S) ([]any, error) {
+	var array []any
```
marcboeker, line 400: nit: `extractSlice()` but with an `array` var assignment is a bit confusing.

marcboeker (approved):
@taniabogatsch Thanks for the type addition. I've added some nits as comment.

rows.go
taniabogatsch: We are returning a slice because we cannot dynamically return differently sized arrays, to the best of my knowledge... Scanning into the composite array type works, but I am not sure how I'd generalize that to return an array in `vector_getters.`

type_info.go
taniabogatsch: This is a duckdb restriction - so I propagated that to Go. 
```
D CREATE TABLE tbl (i INT[0]);
Parser Error: Arrays must have a size of at least 1
```

taniabogatsch:
Hi, @marcboeker - thanks for the review! I've implemented your comments and replied to some others!

marcboeker (approved):

PR #307, opened 2024-11-07
Fix memory leak in prepared statements

begelundmuller:
This PR fixes a potential memory leak where a statement is not closed if `prepared.ExecContext` errors.


marcboeker:
@begelundmuller Thanks for the fix!

PR #308, opened 2024-11-09
add String() method to Decimal

veqryn:
Right now, `Decimal` only has a `Float64` method and an unused `toString()`, which returns just the data type.

The Float64 method is great, but like all floats, it can introduce errors because of how floats work.

Example:
```golang
	// This outputs: 0.008216620000000001
	// It should be: 0.00821662
	fmt.Println((&duckdb.Decimal{
		Width: 18,
		Scale: 8,
		Value: big.NewInt(821662),
	}).Float64())
```

This PR adds a Stinger method allowing a lossless and correct conversion from Decimal to a string.

### DIFF
--- a/types.go
+++ b/types.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math/big"
+	"strings"
 
 	"github.com/mitchellh/mapstructure"
 )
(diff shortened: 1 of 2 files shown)


types.go
```diff
@@ -130,6 +131,36 @@ func (d *Decimal) Float64() float64 {
...
+	// Pad a number with 0.0's if needed
+	if len(zeroTrimmed) <= scale {
+		return fmt.Sprintf("%s0.%s%s", signStr, strings.Repeat("0", scale-len(zeroTrimmed)), zeroTrimmed)
+	}
+	return signStr + zeroTrimmed[:len(zeroTrimmed)-scale] + "." + zeroTrimmed[len(zeroTrimmed)-scale:]
+}
+
 func (d *Decimal) toString() string {
 	return fmt.Sprintf("DECIMAL(%d,%d)", d.Width, d.Scale)
 }
```
taniabogatsch, line 166: Let's remove this. As you pointed out, it is not exported and is unused. 😅

types_test.go
```diff
@@ -381,6 +381,151 @@ func TestDecimal(t *testing.T) {
...
+			},
+			expected: "0.00821662",
+		},
+	}
+
+	for _, tc := range testCases {
+		actual := tc.input.String()
+		if actual != tc.expected {
+			t.Errorf("ToString(%v) => %q, want %q", tc.input, actual, tc.expected)
+		}
```
taniabogatsch, line 525: For consistency with our other tests:
```Go
require.Equal(t, tc.expected, actual)
```

types.go
veqryn: 👍

types_test.go
veqryn: 👍

veqryn:
Thank you @taniabogatsch , should be good now

taniabogatsch:
Thanks!

PR #310, opened 2024-11-12
Expose conn and stmt for advanced usage

fanyang01:
Hi @taniabogatsch,

This PR exposes `conn` and `stmt`, enabling access according to Go’s visibility rules. Additionally, it makes more information about `stmt` accessible.

The motivation for this comes from the [MyDuck Server](https://github.com/apecloud/myduckserver) project, which I’m actively developing. MyDuck Server allows users to run DuckDB as a server that supports MySQL and Postgres wire protocols, leveraging go-duckdb extensively—and it’s been working well so far. Thanks for your continued maintenance!

Currently, I'm implementing prepared statement support for the Postgres protocol, which requires access to parameter and return types for prepared statements (the client sends `Describe` messages to the server to retrieve this info). With the changes in this PR, I would be able to retrieve parameter types using the [Conn.Raw](https://pkg.go.dev/database/sql#Conn.Raw) method.

However, the C API still lacks access to the return types of a prepared statement (though it is available in the C++ API). I’m considering a workaround, like executing the query with `LIMIT 0` to infer return types.

Thanks for considering this PR!

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -15,15 +15,51 @@ import (
 	"unsafe"
 )
 
-type stmt struct {
-	c                *conn
+type StmtType C.duckdb_statement_type
+
+const (
+	DUCKDB_STATEMENT_TYPE_INVALID      StmtType = C.DUCKDB_STATEMENT_TYPE_INVALID
+	DUCKDB_STATEMENT_TYPE_SELECT       StmtType = C.DUCKDB_STATEMENT_TYPE_SELECT
+	DUCKDB_STATEMENT_TYPE_INSERT       StmtType = C.DUCKDB_STATEMENT_TYPE_INSERT
+	DUCKDB_STATEMENT_TYPE_UPDATE       StmtType = C.DUCKDB_STATEMENT_TYPE_UPDATE
+	DUCKDB_STATEMENT_TYPE_EXPLAIN      StmtType = C.DUCKDB_STATEMENT_TYPE_EXPLAIN
+	DUCKDB_STATEMENT_TYPE_DELETE       StmtType = C.DUCKDB_STATEMENT_TYPE_DELETE
+	DUCKDB_STATEMENT_TYPE_PREPARE      StmtType = C.DUCKDB_STATEMENT_TYPE_PREPARE
+	DUCKDB_STATEMENT_TYPE_CREATE       StmtType = C.DUCKDB_STATEMENT_TYPE_CREATE
+	DUCKDB_STATEMENT_TYPE_EXECUTE      StmtType = C.DUCKDB_STATEMENT_TYPE_EXECUTE
+	DUCKDB_STATEMENT_TYPE_ALTER        StmtType = C.DUCKDB_STATEMENT_TYPE_ALTER
+	DUCKDB_STATEMENT_TYPE_TRANSACTION  StmtType = C.DUCKDB_STATEMENT_TYPE_TRANSACTION
+	DUCKDB_STATEMENT_TYPE_COPY         StmtType = C.DUCKDB_STATEMENT_TYPE_COPY
+	DUCKDB_STATEMENT_TYPE_ANALYZE      StmtType = C.DUCKDB_STATEMENT_TYPE_ANALYZE
+	DUCKDB_STATEMENT_TYPE_VARIABLE_SET StmtType = C.DUCKDB_STATEMENT_TYPE_VARIABLE_SET
+	DUCKDB_STATEMENT_TYPE_CREATE_FUNC  StmtType = C.DUCKDB_STATEMENT_TYPE_CREATE_FUNC
+	DUCKDB_STATEMENT_TYPE_DROP         StmtType = C.DUCKDB_STATEMENT_TYPE_DROP
+	DUCKDB_STATEMENT_TYPE_EXPORT       StmtType = C.DUCKDB_STATEMENT_TYPE_EXPORT
+	DUCKDB_STATEMENT_TYPE_PRAGMA       StmtType = C.DUCKDB_STATEMENT_TYPE_PRAGMA
+	DUCKDB_STATEMENT_TYPE_VACUUM       StmtType = C.DUCKDB_STATEMENT_TYPE_VACUUM
+	DUCKDB_STATEMENT_TYPE_CALL         StmtType = C.DUCKDB_STATEMENT_TYPE_CALL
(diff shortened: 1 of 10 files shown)


taniabogatsch:
Hi @fanyang01, thanks for your PR :) I don't quite follow how you call your new functions by exposing the raw driver connection. How do you retrieve the prepared `Stmt`? I.e., could you provide some tests for: 
- `func (s *Stmt) ParamName(n int) string`
- `func (s *Stmt) ParamType(n int) Type`
- `func (s *Stmt) StatementType() StmtType`

That would help me better understand why you want to expose both `Conn` and `Stmt`.

> However, the C API still lacks access to the return types of a prepared statement (though it is available in the C++ API). I’m considering a workaround, like executing the query with LIMIT 0 to infer return types.

PRs against duckdb main are always welcome - you could tag me on it.

taniabogatsch:
In general, I am not opposed to exposing the connection and the statement.
Other database drivers (I checked [go-sqlite3](https://github.com/mattn/go-sqlite3)) seem to do the same.

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
 	return nil
 }
 
-func (s *stmt) NumInput() int {
+// NumInput returns the number of placeholders in the statement.
```
taniabogatsch, line 76: nit: `placeholder parameters`

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
...
 	paramCount := C.duckdb_nparams(*s.stmt)
 	return int(paramCount)
 }
 
-func (s *stmt) bind(args []driver.NamedValue) error {
+// ParamName returns the name of the parameter at the given index (1-based).
+func (s *Stmt) ParamName(n int) string {
+	if s.closed {
+		panic("database/sql/driver: misuse of duckdb driver: ParamName after Close")
+	}
```
taniabogatsch, line 90: Let's return an error instead. In newer code paths, we've been trying to turn some of the `panics` into errors instead. Same for `ParamType` and `StatementType`.

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
...
 		panic("database/sql/driver: misuse of duckdb driver: NumInput after Close")
 	}
 	paramCount := C.duckdb_nparams(*s.stmt)
 	return int(paramCount)
 }
 
-func (s *stmt) bind(args []driver.NamedValue) error {
+// ParamName returns the name of the parameter at the given index (1-based).
+func (s *Stmt) ParamName(n int) string {
+	if s.closed {
```
taniabogatsch, line 88: Let's add a test triggering this error. Does this correctly handle indexes that are out of bounds? I.e., should it return an empty string or an error? Same for `ParamType`.

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
...
+	}
+
+	name := C.duckdb_parameter_name(*s.stmt, C.idx_t(n))
+	paramName := C.GoString(name)
+	C.duckdb_free(unsafe.Pointer(name))
+	return paramName
+}
+
+// ParamType returns the expected type of the parameter at the given index (1-based).
+func (s *Stmt) ParamType(n int) Type {
```
taniabogatsch, line 99: Let's add a test for this error and ensure that we return `TYPE_INVALID` for out-of-bounds indexes.

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
...
 }
 
-func (s *stmt) bind(args []driver.NamedValue) error {
+// ParamName returns the name of the parameter at the given index (1-based).
+func (s *Stmt) ParamName(n int) string {
+	if s.closed {
+		panic("database/sql/driver: misuse of duckdb driver: ParamName after Close")
+	}
+
+	name := C.duckdb_parameter_name(*s.stmt, C.idx_t(n))
```
taniabogatsch, line 92: If we expose the `Stmt`, should we improve error handling when using the API with uninitialized statements? I'd prefer to return an invalid statement type. What do you think?
E.g., this currently panics.
```Go
	var emptyStmt Stmt
	require.Equal(t, DUCKDB_STATEMENT_TYPE_INVALID, emptyStmt.StatementType())
```
Same for `ParamType` and `StatementType`.

statement.go
```diff
@@ -36,15 +73,56 @@ func (s *stmt) Close() error {
...
+
+	return StmtType(C.duckdb_prepared_statement_type(*s.stmt))
+}
+
+// Bind binds the arguments to the query.
+// WARNING: This is a low-level API and should be used with caution.
+func (s *Stmt) Bind(args []driver.NamedValue) error {
+	if s.closed {
+		return errors.Join(errCouldNotBind, errClosedCon)
+	}
```
taniabogatsch, line 121: Missing test.

statement.go
```diff
@@ -191,33 +301,47 @@ func (s *stmt) ExecContext(ctx context.Context, nargs []driver.NamedValue) (driv
...
 	return s.QueryContext(context.Background(), argsToNamedArgs(args))
 }
 
-func (s *stmt) QueryContext(ctx context.Context, nargs []driver.NamedValue) (driver.Rows, error) {
-	res, err := s.execute(ctx, nargs)
-	if err != nil {
+// QueryContext executes a query that may return rows, such as a SELECT.
+// It implements the driver.StmtQueryContext interface.
+func (s *Stmt) QueryContext(ctx context.Context, nargs []driver.NamedValue) (driver.Rows, error) {
+	if err := s.Bind(nargs); err != nil {
```
taniabogatsch, line 311: Am I missing something here? It looks like we changed the behavior of `QueryContext`: Instead of `execute`, it now performs `Bind` and `QueryBound`.

taniabogatsch (changes requested):
Hi, @fanyang01, thanks for adding the tests. I've had another more detailed pass over your changes. Thanks also for adding more comments to the existing functions! 

One thing I was wondering: this PR is becoming quite involved and touches a lot of critical-ish code paths. Would it be possible to split it into a few smaller PRs? E.g., 
1. Expose `Conn` and `Stmt` and the comments.
2. `ParamName`, `ParamType`, and `StatementType` and their tests.
3. New exposed low-level API calls for binding and their tests.

That would make it easier for me to review and know what's happening - no worries if it is too much work. Then, we'll work with this PR. :smiling:

fanyang01:
Hi @taniabogatsch, thank you very much for the detailed review! I will split this PR into several smaller ones, as you suggested, later this week.

fanyang01:
Splitted into #319, #324, and #333

PR #317, opened 2024-11-23
Streaming arrow query results

fanyang01:
This PR updates `Arrow.QueryContext` to return a streaming Arrow record reader, rather than materializing all query results in memory. This approach is beneficial for handling query results larger than available memory, such as when exporting a large table.

### DIFF
--- a/arrow.go
+++ b/arrow.go
@@ -91,6 +91,75 @@ func NewArrowFromConn(driverConn driver.Conn) (*Arrow, error) {
 	return &Arrow{c: dbConn}, nil
 }
 
+// arrowStreamReader implements array.RecordReader for streaming DuckDB results.
+type arrowStreamReader struct {
+	ctx        context.Context
+	stmt       *stmt
+	res        *C.duckdb_arrow
+	schema     *arrow.Schema
+	rowCount   uint64
+	readCount  uint64
+	currentRec arrow.Record
+	err        error
+}
+
+func (r *arrowStreamReader) Retain() {}
+
+func (r *arrowStreamReader) Release() {
+	if r.currentRec != nil {
+		r.currentRec.Release()
+		r.currentRec = nil
+	}
+	if r.res != nil {
+		C.duckdb_destroy_arrow(r.res)
+		r.res = nil
+	}
+	if r.stmt != nil {
+		r.stmt.Close()
+		r.stmt = nil
+	}
+}
+
+func (r *arrowStreamReader) Schema() *arrow.Schema {
+	return r.schema
+}
+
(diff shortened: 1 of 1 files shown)


arrow.go
```diff
@@ -91,6 +91,75 @@ func NewArrowFromConn(driverConn driver.Conn) (*Arrow, error) {
...
+	readCount  uint64
+	currentRec arrow.Record
+	err        error
+}
+
+func (r *arrowStreamReader) Retain() {}
+
+func (r *arrowStreamReader) Release() {
+	if r.currentRec != nil {
+		r.currentRec.Release()
```
Tishj, line 110: If this is a `release` callback for the ArrowArray, it should check whether `release` is null before calling it
> A released structure is indicated by setting its release callback to NULL. Before reading and interpreting a structure’s data, consumers SHOULD check for a NULL release callback and treat it accordingly (probably by erroring out).

arrow.go
```diff
@@ -91,6 +91,75 @@ func NewArrowFromConn(driverConn driver.Conn) (*Arrow, error) {
...
+	schema     *arrow.Schema
+	rowCount   uint64
+	readCount  uint64
+	currentRec arrow.Record
+	err        error
+}
+
+func (r *arrowStreamReader) Retain() {}
+
+func (r *arrowStreamReader) Release() {
```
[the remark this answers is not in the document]
fanyang01: Hi @Tishj,  

Thank you very much for the review! It seems that Go's Arrow binding wraps a Go `RecordReader` into an `ArrowArrayStream` and conforms to the spec with a release wrapper: https://github.com/apache/arrow-go/blob/v18.1.0/arrow/cdata/exports.go#L175  

I’ve also updated the implementation to include an atomic reference count, similar to how the official `RecordReader` is implemented: https://github.com/apache/arrow-go/blob/v18.1.0/arrow/array/record.go#L44

arrow.go
fanyang01: Unfortunately, this is not a callback but a method, so it cannot be null.

taniabogatsch (commented):
Hi, @fanyang01, thanks for implementing @Tishj's suggestions! 
I also just looked at your PR - should there be tests addressing/covering the new (exported) functionality?

datbth:
hi @fanyang01, thank you for working on this.  
Will you be able to continue working on this soon, or would you mind if I branch out from yours and make a new PR?

taniabogatsch:
@datbth - I am not @fanyang01, but from my side you're welcome to open a follow-up PR based on this PR so we can get this ready to go in. :)

PR #319, opened 2024-11-25
Expose conn and stmt [Part 1]

fanyang01:
This PR is Part 1 of #310. It renames `conn` to `Conn` and `stmt` to `Stmt`. Additionally, comments have been added for all public methods. No other changes are included.


taniabogatsch:
Thanks!

PR #324, opened 2024-11-27
Expose statement type, parameter type, and parameter name

fanyang01:
This PR is part 2 of #310. Based on the code review feedback for #310, error handling has been improved by checking for uninitialized `Stmt`, detecting out-of-bounds indices, and reporting closed statements.

@taniabogatsch

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -15,6 +15,39 @@ import (
 	"unsafe"
 )
 
+type StmtType C.duckdb_statement_type
+
+const (
+	DUCKDB_STATEMENT_TYPE_INVALID      StmtType = C.DUCKDB_STATEMENT_TYPE_INVALID
+	DUCKDB_STATEMENT_TYPE_SELECT       StmtType = C.DUCKDB_STATEMENT_TYPE_SELECT
+	DUCKDB_STATEMENT_TYPE_INSERT       StmtType = C.DUCKDB_STATEMENT_TYPE_INSERT
+	DUCKDB_STATEMENT_TYPE_UPDATE       StmtType = C.DUCKDB_STATEMENT_TYPE_UPDATE
+	DUCKDB_STATEMENT_TYPE_EXPLAIN      StmtType = C.DUCKDB_STATEMENT_TYPE_EXPLAIN
+	DUCKDB_STATEMENT_TYPE_DELETE       StmtType = C.DUCKDB_STATEMENT_TYPE_DELETE
+	DUCKDB_STATEMENT_TYPE_PREPARE      StmtType = C.DUCKDB_STATEMENT_TYPE_PREPARE
+	DUCKDB_STATEMENT_TYPE_CREATE       StmtType = C.DUCKDB_STATEMENT_TYPE_CREATE
+	DUCKDB_STATEMENT_TYPE_EXECUTE      StmtType = C.DUCKDB_STATEMENT_TYPE_EXECUTE
+	DUCKDB_STATEMENT_TYPE_ALTER        StmtType = C.DUCKDB_STATEMENT_TYPE_ALTER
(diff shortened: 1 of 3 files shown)


statement.go
```diff
@@ -15,6 +15,39 @@ import (
 	"unsafe"
 )
 
+type StmtType C.duckdb_statement_type
+
+const (
+	DUCKDB_STATEMENT_TYPE_INVALID      StmtType = C.DUCKDB_STATEMENT_TYPE_INVALID
```
taniabogatsch, line 21: Let's make this consistent with the types in `type.go`. I.e, let's not repeat the package name `DUCKDB_`:
```Go
STATEMENT_TYPE_INVALID      StmtType = C.DUCKDB_STATEMENT_TYPE_INVALID
```

statement.go
```diff
@@ -49,6 +82,55 @@ func (s *Stmt) NumInput() int {
 	return int(paramCount)
 }
 
+// ParamName returns the name of the parameter at the given index (1-based).
+func (s *Stmt) ParamName(n int) (string, error) {
+	if s.closed {
+		return "", errClosedStmt
```
taniabogatsch, line 88: Since these are helper functions, as in, they are not part of Go's sql driver interface, we might want to wrap their errors in an API error, like we do for `TypeInfo`.
```Go
func NewTypeInfo(t Type) (TypeInfo, error) {
	name, inMap := unsupportedTypeToStringMap[t]
	if inMap && t != TYPE_ANY {
		return nil, getError(errAPI, unsupportedTypeError(name))
	}
// ...
```

statement_test.go
```diff
@@ -73,9 +184,104 @@ func TestPrepareQueryNamed(t *testing.T) {
...
+		// require.Equal(t, TYPE_VARCHAR, paramType)
+		require.Equal(t, TYPE_INVALID, paramType)
+		require.NoError(t, err)
+
+		paramType, err = stmt.ParamType(2)
+		// Will be fixed in the next release.
+		// https://github.com/duckdb/duckdb/pull/14952
+		// require.Equal(t, TYPE_INTEGER, paramType)
+		require.Equal(t, TYPE_INVALID, paramType)
+		require.NoError(t, err)
```
taniabogatsch, line 242: Let's put a `FIXME` here with a link to an issue (can you open one?). So we don't forget to come back to this.

taniabogatsch (changes requested):
Hey @fanyang01, thanks for the second part of your original PR! :) I just had a look and only left a few nits w.r.t. the error handling. Otherwise, this should be good to go.

statement.go
fanyang01: Done

statement.go
fanyang01: Done, although I don't fully understand why we should wrap the error, as the true error will be suppressed by the `getError` function. This formatting with `%s` prevents us from using `errors.Is` in the tests.

statement_test.go
fanyang01: `require.Equal(t, TYPE_INVALID, paramType)` will fail when upgrading to the new DuckDB version, ensuring that we won't overlook this.

fanyang01:
@taniabogatsch Thank you very much for the detailed review :) I have updated the PR.

statement.go
taniabogatsch: Ah, this is indeed a shortcoming of `getError`... It should probably use [`errors.Join`](https://cs.opensource.google/go/go/+/go1.23.4:src/errors/join.go;l=19). Thanks for pointing that out! Would you prefer we keep the original error (until this is addressed in `getError`)?

fanyang01: Yes, `errors.Join` will solve the problem. I have reverted the error-wrapping changes.

taniabogatsch (approved):

PR #325, opened 2024-11-27
Getters and setters for the JSON type

taniabogatsch:
cc @marcboeker, this might break some backward compatibility, as we no longer assume that every `JSON` column scan returns a `string.` Instead, we return the respective JSON interface. This aligns with the duckdb CLI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,10 @@ install:
 
 .PHONY: examples
 examples:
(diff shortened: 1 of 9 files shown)


marcboeker:
Thanks! LGTM. Should we document this somewhere or add a JSON example?

marcboeker:
Thanks!

PR #327, opened 2024-12-09
Fix UUID.Scan

apocelipes:
Fix:
- db.QureyRow(`select 'urn:uuid:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'`).Scan(&uuid) panics, because the value type is string. The string value is a valid uuid text.
- db.QureyRow(`select more-than-16-bytes-and-invalid-blob`).Scan(&uuid), this should return an error, but it reports no error and creates a wrong uuid value.

Add a `String()` method to UUID, it encodes the uuid value to a string.

### DIFF
--- a/types.go
+++ b/types.go
@@ -38,7 +38,7 @@ func (u *UUID) Scan(v any) error {
 		}
 		copy(u[:], id[:])
 	default:
-		return fmt.Errorf("invalid UUID type: %#v", val)
+		return fmt.Errorf("invalid UUID value type: %T", val)
 	}
 	return nil
 }
--- a/types.go
+++ b/types.go
@@ -7,10 +7,12 @@ import "C"
 
 import (
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"strings"
 
+	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


types.go
```diff
@@ -23,12 +25,40 @@ const uuid_length = 16
...
+		copy(u[:], val[:])
+	case string:
+		id, err := uuid.Parse(val)
+		if err != nil {
+			return err
+		}
+		copy(u[:], id[:])
+	default:
+		return fmt.Errorf("invalid UUID type: %#v", val)
 	}
```
taniabogatsch, line 42: Thanks for the changes! Do we already have tests covering these (error) code paths?

taniabogatsch (commented):
Hi @apocelipes, thanks for fixing this!

types.go
apocelipes: Not yet, but I can add one test case to do that.

taniabogatsch: Perfect! :)

apocelipes: All three error paths has been tested (type check, []byte values check and string values check). You could take a look.

taniabogatsch:
Thanks!

PR #333, opened 2024-12-16
Expose `Bind` and `(Query|Exec)Bound` on `Stmt` for advanced usage

fanyang01:
This is part 3 of #310. Based on the original review feedback, the existing `(Query|Exec)Context` APIs remain unchanged and will continue to panic when called on a closed or ongoing statement. In contrast, the three newly introduced APIs are implemented to return errors in such scenarios.

The newly introduced APIs enable developers to bridge the prepared statement mechanisms between DuckDB and other database protocols, such as PostgreSQL and MySQL.

Looking forward to your feedback, @taniabogatsch. Thanks!


errors.go
```diff
@@ -106,6 +106,9 @@ var (
 	errPrepare                    = errors.New("could not prepare query")
 	errMissingPrepareContext      = errors.New("missing context for multi-statement query: try using PrepareContext")
 	errEmptyQuery                 = errors.New("empty query")
+	errCouldNotBind               = errors.New("could not bind parameter")
+	errActiveRows                 = errors.New("ExecContext or QueryContext with active Rows")
```
taniabogatsch, line 110: Could we try to trigger this error in the tests?

taniabogatsch (commented):
Hi, @fanyang01 - thanks for staging the PRs! Looks great - just a small comment about a missing test.

errors.go
fanyang01: Done. Thanks for spotting this!

taniabogatsch:
Great, it looks good now! Thanks for the PRs in this series. :)

Issue #305, opened 2024-11-07
No results for UNION type


jorinvo:
Hi and thanks for this driver. It works really well until I ran into the following:

A query that contains a union type doesn't return any rows.

In the CLI I can do this:

```
$ duckdb 
v1.1.2 f680b7d08f
Enter ".help" for usage hints.
D CREATE TYPE MYUNION AS UNION(s VARCHAR);
D SELECT 'test'::MYUNION;
┌─────────────────────────┐
│ CAST('test' AS MYUNION) │
│    union(s varchar)     │
├─────────────────────────┤
│ test                    │
└─────────────────────────┘
```

But in Go the second fmt will never show up:

```go
func main() {
	db, err := sql.Open("duckdb", "")
	if err != nil {
		panic(err)
	}
	_, err = db.Exec("CREATE TYPE MYUNION AS UNION(s VARCHAR);")
	if err != nil {
		panic(err)
	}
	rows, err := db.Query("SELECT 'test';")
	for rows.Next() {
		var s string
		rows.Scan(&s)
		fmt.Println("this works: ", s)
	}
	rows, err = db.Query("SELECT 'test'::MYUNION;")
	for rows.Next() {
		var s string
		rows.Scan(&s)
		fmt.Println("this does not: ", s)
	}
	defer db.Close()
}
```

I originally ran into this in a project and could replicate it isolated like this.
Any ideas what is going on here?
Thanks a lot!

taniabogatsch:
Hi @jorinvo! This is likely because we do not yet support the `UNION` type in this driver. However, I would have expected an error (unsupported type) message instead of an empty result.

In https://github.com/marcboeker/go-duckdb/blob/a32843693a7b014008d6514a13dcda3e13dccb99/type.go#L51C1-L60C2.
```Go
// FIXME: Implement support for these types.
var unsupportedTypeToStringMap = map[Type]string{
	TYPE_INVALID:  "INVALID",
	TYPE_UHUGEINT: "UHUGEINT",
	TYPE_ARRAY:    "ARRAY",
	TYPE_UNION:    "UNION",
	TYPE_BIT:      "BIT",
	TYPE_ANY:      "ANY",
	TYPE_VARINT:   "VARINT",
}
```

I'm currently implementing the `ARRAY` type. I can look afterward and either add the expected error or implement `UNION` support (if you don't want to pick that up yourself 😄 ).

jorinvo:
[...]

PR #356, opened 2025-01-28
Add support for UNION types

jorinvo:
This addresses #305. 
~I only implemented reading union types since that is all I need for the time being.~
The PR implements read+write support for Union types.

- Database type name is returned in the SQL string format like `UNION(num INTEGER, str VARCHAR)`
- Scan type is ~interface{}~ `duckdb.Union`
- Writing should work with `duckdb.Union` types or passing the underlying member types and letting the driver pick the first matching one.


jorinvo:
Seems like a lot has happened in the 1.5 months since I opended the PR.
As far as I can tell the bindings are in a [new repo](https://github.com/duckdb/duckdb-go-bindings/) now. But there are no bindings for union types yet.
I don't know if I should attempt to create those in the repo over there or what your plan is around this.
I just don't want to redo all the work without knowing if I am even doing the right thing or if it will conflict with the next refactoring and I have to redo the PR once more.

It would be awesome if you could check the PR and let me if it's fine in general and if you intend to merge it once I redo it on top of the new internal mappings API. I understand if you prefer a different approach for union types. And please tell me if there are any other refactorings planned that would conflict with this. And let me know if I should go ahead and create a PR on the bindings repo and then update this PR here accordingly.

I hope you understand my confusion and thank you.

taniabogatsch:
Yes, I understand. These are very valid points. Sorry for the confusion! We had to find a solution for the binaries, as they started exceeding the GitHub file size limit starting at duckdb 1.2.0. We decided to encapsulate them in their module. Hopefully, no other big refactoring will be necessary soon. 😅 

So yes, that module is very new, and indeed, as you've pointed out, it is still missing many of the C API functions. I'll see if I can push more of them this week and then ping you. I'll also tag you on the PR adding the bindings, to make sure that I don't miss any you'll need.

> It would be awesome if you could check the PR and let me if it's fine in general and if you intend to merge it once I redo it on top of the new internal mappings API.

From what I can tell after scrolling through the changes, the PR looks great, pretty much like I've invisioned it! Maybe we want to expose `driver.Value` instead of `interface`, and I might have a few more comments around it, once I give it an in-depth review. But it definitely looks like something I intend to merge. :)

jorinvo:
That sounds great! Thanks for the quick feedback @taniabogatsch. Looking forward to getting this wrapped up and merged :slightly_smiling_face:

jorinvo:
Thanks @taniabogatsch! I see if I can get it to work now.

jorinvo:
@taniabogatsch I resolved the conflicts and reading union types seems to be genrally working also with v2 now.

But I encountered some issues with the current concepts of using `interface{}` as type:
- Using my fork for the last few months I noticed that detecing the underlying type the returned union value is difficult. When getting a value of type `interface{}` you have to try out different Go types to find the correct type. But SQL TIME and TIMESTAMP both result in a time.Time. So I have to look at the date part and to guess which underlying type was used. There are probably more things I haven't run into yet.
- I tried using `duckdb.Value` instead of `interface{}`, but the scanType for `duckdb.Value` is also `interface{}`.
- I encountered issues reflect.TypeOf when trying to add more testscases in duckdb_test.go.
- I also tried implementing writing union types and encountered issues around this.

What do you think of representing union types in Golang as a new type similar to how we already have `duckdb.Interval` and so on?

I would imagine something like:

```go
type Union struct {
  Tag string
  Value Value
}
```

This would expose the internal member name as tag similarly how it is [available via SQL](https://duckdb.org/docs/stable/sql/functions/union).
What do you think about this?

taniabogatsch:
Indeed, we've been using light wrappers for types like `Interval` and `Decimal` already. I think the main issue with adding and exposing these custom types is that we are implementing our own type wrapper module (inside go-duckdb). There's a PR [here](https://github.com/marcboeker/go-duckdb/pull/365) to create a better separation.

That being said, I agree that for `UNION` we need a custom type, too, as we need to expose the `tag` as `string`. We already expose the `TypeInfo` and `Type` interface, so we might also want to consider adding that to the `Union` struct.

What is `Value` in your example? Because for performance reasons I think we want to stay away from the C API's `Value` interface. Would `any` in combination with `TypeInfo` work? That should allow people to switch on the type?

jorinvo:
Sounds good.
I will adjust the PR.

With `Value` I meant (SQL) `driver.Value` (which is defined as `any`. Might be less confusing to just use `any` directly.

It makes sense to expose the type of the union member. If I am understand you correctly you are referring to `duckdb.Type` that is also exposed via `TypeInfo.InternalType()`, right? This `Type` internally points to `mapping.Type`, which points to `binding.Type`, right?

Does the following adjustment to the proposed struct make sense:

```go
type Union struct {
  Value any
  Tag   string
  Type  Type
}
```

taniabogatsch:
> With Value I meant (SQL) `driver.Value`

Ah, I think `driver.Value` is actually more descriptive than just `any`, so that's cool.

Hmm, now that I think more about it,`Type` misses some information for nested types, e.g., if `Type` is `TYPE_LIST`, it is not possible to know the child type of the list... But exposing `TypeInfo` also feels incomplete, as there are no functions to gain child types, etc. yet. 

On second thought, maybe we only expose `{ Value driver.Value Tag string}` for now, allowing us to add additional struct members in a separate PR (if necessary). In theory, the user should know the type of a `Tag`. What do you think?

taniabogatsch:
Great, sounds good! :)

jorinvo:
I update the PR. It's ready for review now. Let me know if there is anything we should add or change.

scalar_udf.go
taniabogatsch, line 1: Do we need to add the same "not implemented" checks to the table UDF code paths?

vector.go
```diff
@@ -134,7 +136,9 @@ func (vec *vector) getChildVectors(v mapping.Vector, writable bool) {
 	switch vec.Type {
 	case TYPE_LIST, TYPE_MAP:
 		child := mapping.ListVectorGetChild(v)
-		vec.childVectors[0].initVectors(child, writable)
+		if child.Ptr != nil {
+			vec.childVectors[0].initVectors(child, writable)
+		}
```
taniabogatsch, line 141: How do you trigger `child.Ptr` being `nil`? That seems like unintentional behaviour?

errors_test.go
```diff
@@ -485,3 +485,94 @@ func TestGetDuckDBErrorIs(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// Test inserting wrong type
+		_, err = db.Exec(`
+            INSERT INTO test_union VALUES ((true)::UNION(num INTEGER, str VARCHAR))
+        `)
+		var de *Error
+		require.True(t, errors.As(err, &de))
+		require.Equal(t, ErrorTypeConversion, de.Type)
+	})
```
taniabogatsch, line 544: These are DuckDB errors and not necessarily related to go-duckdb's `UNION` implementation. I don't think we necessarily need them. Same for `"invalid member type"` and `"missing member name"`.

errors_test.go
```diff
@@ -485,3 +485,94 @@ func TestGetDuckDBErrorIs(t *testing.T) {
...
+		require.NoError(t, err)
+		defer rows.Close()
+
+		require.True(t, rows.Next())
+		var nullUnion Union
+		err = rows.Scan(&nullUnion)
+		require.NoError(t, err)
+		require.Equal(t, "", nullUnion.Tag)
+		require.Nil(t, nullUnion.Value)
+	})
```
taniabogatsch, line 559: Can we move this test to `types_test`, as it strictly speaking does not trigger any `UNION`-related error message?

vector.go
```diff
@@ -507,6 +518,35 @@ func (vec *vector) initArray(logicalType mapping.LogicalType, colIdx int) error
...
+
+		err := vec.childVectors[i].init(memberType, colIdx)
+		mapping.DestroyLogicalType(&memberType)
+		if err != nil {
+			return err
+		}
+	}
+
+	vec.Type = TYPE_UNION
+	vec.getFn = func(vec *vector, rowIdx mapping.IdxT) any {
```
taniabogatsch, line 540: Let's also add a "not implemented/ invalid code path" `panic` to `vec.setFn` (a dummy callback). That callback should be unreachable code. However, currently it is possible to trigger a panic/crash by trying to append to a `UNION` column (with the test below). Instead, we need to throw a `not implemented` there. Or, maybe, we also add support for `setFn` to this PR?

```Go
func TestAppenderUnion(t *testing.T) {
	c, db, conn, a := prepareAppender(t, `
	CREATE TABLE test (
		i UNION(num INTEGER, str VARCHAR)
	)`)
	defer cleanupAppender(t, c, db, conn, a)

	union := Union{
		Tag:   "num",
		Value: int32(42),
	}

	require.NoError(t, a.AppendRow(union))
	require.NoError(t, a.Flush())
}
```

vector_getters.go
```diff
@@ -201,3 +201,49 @@ func (vec *vector) getSliceChild(offset uint64, length uint64) []any {
 	}
 	return slice
 }
+
+func (vec *vector) getUnion(rowIdx mapping.IdxT) any {
+	if vec.getNull(rowIdx) {
+		return Union{Tag: "", Value: nil}
+	}
```
taniabogatsch, line 208: Removable?

vector_getters.go
```diff
@@ -201,3 +201,49 @@ func (vec *vector) getSliceChild(offset uint64, length uint64) []any {
...
+
+	// Get the tag vector (stored at index 0)
+	tagVector := mapping.StructVectorGetChild(vec.vec, 0)
+	if tagVector.Ptr == nil {
+		return Union{Tag: "", Value: nil}
+	}
```
taniabogatsch, line 214: We should remove all `.Ptr == nil` checks from this function - these cases should never happen, and if they do, they are a bug.

vector_getters.go
```diff
@@ -201,3 +201,49 @@ func (vec *vector) getSliceChild(offset uint64, length uint64) []any {
...
+
+	// Get the tag value
+	tagData := mapping.VectorGetData(tagVector)
+	if tagData == nil {
+		return Union{Tag: "", Value: nil}
+	}
```
taniabogatsch, line 220: Same here, `tagData` should never be `nil`, no?

vector_getters.go
```diff
@@ -201,3 +201,49 @@ func (vec *vector) getSliceChild(offset uint64, length uint64) []any {
...
 	return slice
 }
+
+func (vec *vector) getUnion(rowIdx mapping.IdxT) any {
+	if vec.getNull(rowIdx) {
+		return Union{Tag: "", Value: nil}
+	}
+
+	// Get the tag vector (stored at index 0)
+	tagVector := mapping.StructVectorGetChild(vec.vec, 0)
```
taniabogatsch, line 211: For performance, maybe instead of retrieving the tag vector on every row that's scanned, we should cache it as part of the child vectors? Then, we can also cache its data pointer, instead of calling `VectorGetData` on each row being scanned.

vector_getters.go
```diff
@@ -201,3 +201,49 @@ func (vec *vector) getSliceChild(offset uint64, length uint64) []any {
...
+	tempVec := vector{
+		vec:            memberVector,
+		dataPtr:        mapping.VectorGetData(memberVector),
+		maskPtr:        mapping.VectorGetValidity(memberVector),
+		vectorTypeInfo: vec.childVectors[tag].vectorTypeInfo,
+		getFn:          vec.childVectors[tag].getFn,
+		childVectors:   vec.childVectors[tag].childVectors,
+	}
+
+	value := tempVec.getFn(&tempVec, rowIdx)
```
taniabogatsch, line 244: The struct's child vectors are already cached, we should access them directly via `tag`, which is a `int8t` index, no?

Then, I think we can rewrite all this to:
```Go
	child := &vec.childVectors[tag]
	value := child.getFn(child, rowIdx)
	return Union{
		Value: value,
		Tag:   child.indexDict[tag],
	}
```

`child.name` needs to be initialised during `init` - we need to cache the names for each tag (by index), maybe we can add another dictionary for that
```Go
type vectorTypeInfo struct {
	baseTypeInfo
	namesDict map[string]uint32
   indexDict map[uint32]string
}
```

taniabogatsch (changes requested):
Hi @jorinvo, thanks again for all your work on this PR! I've left a detailed review now :) It looks already really good, I think a few more changes and we can merge this!

type_info.go
```diff
@@ -250,6 +253,38 @@ func NewArrayInfo(childInfo TypeInfo, size uint64) (TypeInfo, error) {
...
+			return nil, getError(errAPI, duplicateNameError(name))
+		}
+		m[name] = true
+	}
+
+	info := &typeInfo{
+		baseTypeInfo: baseTypeInfo{Type: TYPE_UNION},
+		childTypes:   memberTypes,
+		// Store member names
+		enumNames: memberNames, // We can reuse the enumNames field for union member names
```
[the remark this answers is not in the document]
jorinvo: Good point. Updated the names.

vector.go
jorinvo: Good point. I removed the nil checks.

errors_test.go
jorinvo: Good point. Thanks.

errors_test.go
jorinvo: Moved it.

vector_getters.go
jorinvo: Yes, thanks.

vector_getters.go
jorinvo: Thanks' for seeing that.

vector_getters.go
jorinvo: Removed the if

vector_getters.go
jorinvo: Updated the code. The tests are passing. Would be great if you could check again if the implementation makes sense. I have to admit I don't really understand what I am doing here and how data is represented internally in DuckDB. Thank you for bearing with me :pray:

vector_getters.go
jorinvo: That sounds like a good idea. I tried implementing it and the tests are passing.
I added `tagDataPtr unsafe.Pointer` to `vectorTypeInfo` and I updated `getUnion` and `getChildVectors`. Would be great if you can check if I am missing anything. Thanks.

vector.go
jorinvo: I agree, that it feels easier - or at least not much more work to just implement setting union types in the same PR. I gave it a try and implemented `setUnion`.

scalar_udf.go
jorinvo: As I decided it would be easier to implement write support in the same PR, I also adjusted scalar + table UDFs and added tests for that. Hope that's alright. Would be great if you could take a look.

jorinvo:
Hi @taniabogatsch, thank you for the thourough review! It was very much needed :pray: 
I think I addressed all comments. And we should also have write support with this now.
I will adjust the PR title accordingly.
Did I miss anything to fully implement write support?
Please let me know if there is anything else to do and thank again!

jorinvo:
Somehow CI is failing now with segfaults for mac and windows for various tests.
Not sure if this is related to my changes or to me merging the latest main branch into my PR.
Locally on my Linux machine tests are passing. And in CI Ubuntu seems fine too.
Is this a known issue?

scalar_udf.go
taniabogatsch: That's great! I'll have a look now!

vector_getters.go
taniabogatsch: The `UNION` type is a logical type in DuckDB. That means its underlying 'physical type' has to be one of our existing physical types. We chose the `STRUCT` type - it has always as many children as there are tags in the union, plus one. That additional vector (the first child) is a `uint8` vector, and it contains the active tag of the row. Hope that helps. :)

taniabogatsch:
Hi @jorinvo - for a review / feedback, see https://github.com/taleshape-com/go-duckdb/pull/1. :))

vector_getters.go
jorinvo: Thank you for explaining! And thanks for cleaning up the code around this :pray:

taniabogatsch:
Merged - thanks!

jorinvo:
Thanks so much for your support and patience with this PR @taniabogatsch! Feels great seeing it finally merged :partying_face:

taniabogatsch:
Haha of course! - it's always great to collaborate :))

Issue #323, opened 2024-11-26
Statically Iink 'icu' extension by default


wearpants:
For parallel functionality with the [Python bindings & CLI](https://duckdb.org/docs/extensions/core_extensions.html#default-extensions), it'd be helpful to include the `icu` extension in the Makefile, as was done in #280 for JSON.

This enables proper timezone support, and is useful for the [grafana duckdb support](https://github.com/snakedotdev/grafana-duckdb-datasource/) I'm building

taniabogatsch:
Hi, @wearpants - I agree this would align with Python (which has even more) and the homebrew CLI. However, we're already running into GitHub binary size issues, so I hesitate to include more extensions in the pre-build binaries.

wearpants:
Not sure which binary sizes you're referring to, as this project doesn't publish releases on GH…? the lack of parity with CLI/other languages was surprising and the source of bugs that took a few days to track down. The ICU extension is needed not only for time zones but also date time arithmetic

Would really prefer not to be building my own bindings, but if necessary at least having easy Makefile flags for the common modules would be a step in the right direction…

wearpants:
To further clarify my difficulties, I'm building a docker image for grafana with the above duckdb data source plug-in I wrote, and manually installing extensions needs to happen at image build time via the CLI (which is otherwise not present in the image)… All in all quite a hassle and effort would need to be repeated by anyone in a similar environment

taniabogatsch:
> Not sure which binary sizes you're referring to, as this project doesn't publish releases on GH…?

The bundled library files in `deps/...`. We are considering building with `SMALLER_BINARY` or LFS once we move to duckdb's latest version; see here: https://github.com/marcboeker/go-duckdb/issues/336. We hit this with freeBSD and duckdb's latest `main,` the other pre-build libraries are also growing.

> at least having easy Makefile flags for the common modules would be a step in the right direction

I'm happy to review a PR with `Makefile` targets that align with our Homebrew(?) CLI or Python.

taniabogatsch:
If we start publishing `.a` libraries through duckdb, then they'll follow our `.github/config/bundled_extensions.cmake` and have ICU build-in. Linking https://github.com/marcboeker/go-duckdb/discussions/349.

taniabogatsch:
go-duckdb's latest version imports the duckdb-go-bindings, which bundle the default extensions:

duckdb `bundled_extensions.cmake`.
```
#
## Extensions that are linked
#
[...]

PR #351, opened 2025-01-17
Improve appender speed

JAicewizard:
Rough performance before:
```
BenchmarkAppenderSingle/int8-16         	    8288	    142109 ns/op
```

Rough performance after caching chunk size:
```
BenchmarkAppenderSingle/int8-16         	   35270	     32326 ns/op
```

Full (existing API without reusing the appender for fairness) vs the new TS (`TableSource`) implementation
```
BenchmarkAppenderSingleFull/int8-16         	   11076	    109986 ns/op
BenchmarkAppenderSingleFull/int8-16         	   10000	    103503 ns/op
BenchmarkAppenderSingleFull/int8-16         	   10000	    103835 ns/op
BenchmarkAppenderSingleFull/int8-16         	   11526	    107647 ns/op
BenchmarkAppenderSingleFull/int8-16         	   10000	    107233 ns/op
BenchmarkAppenderSingleTS/int8-16           	   12663	     99329 ns/op
BenchmarkAppenderSingleTS/int8-16           	   12016	     97747 ns/op
BenchmarkAppenderSingleTS/int8-16           	   12054	     98896 ns/op
BenchmarkAppenderSingleTS/int8-16           	   12721	     96048 ns/op
BenchmarkAppenderSingleTS/int8-16           	   12434	    102425 ns/op
```

This benchmark only tests the single threaded row based tablesource, if this gets looked at and there are no major problems I can implement the multi-threaded and chunk APIs as well. These will be even faster!

PR #364, opened 2025-02-12
DuckDB v1.2.0 and removing any duckdb.h dependencies

taniabogatsch:
### Breaking changes
```diff
- This PR drops support for FreeBSD, as duckdb does not build and publish any artifacts.
- This PR makes the arrow dependency opt-in.
```

This PR significantly changes go-duckdb's internals and causes a few breaking changes, as outlined above.
Any `C.duckdb...` call has been moved to call into https://github.com/duckdb/duckdb-go-bindings.

### Open ToDo's
- [x] Guide on building and linking a custom FreeBSD static library based on the duckdb code.
- [x] Make `Arrow` opt-in.
- [x] Test duckdb-go-bindings with a custom static/dynamic lib.
- [x] Test go-duckdb with a custom static/dynamic lib.
- [x] CI run to duplicate `mapping.go` for each pre-built combination.
- [x] CI run to test linux_arm64
- [x] Run Instruments with Leak detection
- [x] Update `README`

### Future Work
- [ ] Remove any remaining C imports in the core files.
- [ ] Provide a flag to build this project without importing `"C"`.

### Related Issues / Discussions
- https://github.com/marcboeker/go-duckdb/issues/323
- https://github.com/marcboeker/go-duckdb/discussions/349
- https://github.com/marcboeker/go-duckdb/discussions/232


taniabogatsch:
@marcboeker - this is ready to go in from my side. :)

marcboeker:
@taniabogatsch Awesome. I'll have a look at it. But as the PR is quite large, it could take some time. Maybe we should also get someone else as a second reviewer? Maybe @begelundmuller or @JAicewizard?

taniabogatsch:
Hi, @marcboeker. Yes, it looks like quite the wall. I'll write up a bit to help with the review(s). It seems much bigger than it is.

As I outlined above, this PR moves all `C.duckdb...` calls to instead call into [duckdb-go-bindings](https://github.com/duckdb/duckdb-go-bindings). That means any place where we would previously use `C.duckdb...`, we're now using `apiMyFunc...`.

We cannot directly use `bindings.MyFunc...`, we want to stay as backward compatible as possible. duckdb-go-bindings provides some pre-built OS+architecture combinations, and if we move these imports into so-called 'mapping' files and wrap the function calls into our own `api...` calls, we can continue to ship pre-built versions ourselves.

That means there is quite some duplication around these mapping files. The following are all duplicates; they are just using different import paths. I've also added CI runs/ Makefile targets to perform that duplication automatically. Maybe we can make this nicer with `go generate` in the future.
- Duplicate files: `mapping.go`, `mapping_darwin...`, ...
- Duplicate files for arrow, now opt-in: `arrow_mapping.go`, `arrow_mapping_darwin...`, ...
- Tooling around these: `duplicate_mappings.yml`, `duplicate_arrow_mapping.yml`

Additionally, everything in `deps/...` and the `duckdb.h` has been thrown out, including the custom-built targets, as they're somewhat outdated regarding building duckdb.

The changes to the tests are also mostly aesthetic; I've moved many common functions into wrappers to reduce the code a bit. I started doing this because duckdb-go-bindings allows a 'debug' mode that tracks allocations. With that enabled, it is possible to find resource leaks quite easily (`go test -v -tags=debug_bindings`, so I stepped through all the tests doing that, and fixing some memory leaks.

TLDR: reviewing some of the infrastructure changes in this PR and the actual `.go` files (but not really the mapping files) should reduce the size of the review by quite a bit. :)) Looking forward to your feedback!

data_chunk.go
```diff
@@ -69,10 +64,10 @@ func SetChunkValue[T any](chunk DataChunk, colIdx int, rowIdx int, val T) error
 	if colIdx >= len(chunk.columns) {
 		return getError(errAPI, columnCountError(colIdx, len(chunk.columns)))
 	}
-	return setVectorVal(&chunk.columns[colIdx], C.idx_t(rowIdx), val)
+	return setVectorVal(&chunk.columns[colIdx], uint64(rowIdx), val)
 }
 
-func (chunk *DataChunk) initFromTypes(ptr unsafe.Pointer, types []C.duckdb_logical_type, writable bool) error {
+func (chunk *DataChunk) initFromTypes(types []apiLogicalType, writable bool) error {
```
JAicewizard, line 70: Nice, I really disliked the 2 lists

JAicewizard (commented):
Github seems to once again be broken, so I can only view 10 of the files. Overall My question is, where are all the types defined? I was under the impression they would probably be defined in the duckdb-go-mappings package, but that is never imported from what I can see

rows.go
```diff
@@ -13,38 +8,39 @@ import (
...
 	// chunk holds the currently active data chunk.
 	chunk DataChunk
+	// closeChunk is true after the first iteration of Next.
+	closeChunk bool
 	// chunkCount is the number of chunks in the result.
-	chunkCount C.idx_t
+	chunkCount uint64
 	// chunkIdx is the chunk index in the result.
-	chunkIdx C.idx_t
+	chunkIdx uint64
```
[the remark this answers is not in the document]
taniabogatsch: Sure, I'll expose this in the bindings 👍

rows.go
```diff
@@ -13,38 +8,39 @@ import (
...
 type rows struct {
 	// stmt is a pointer to the stmt of which we are scanning the result.
 	stmt *Stmt
 	// res is the result of stmt.
-	res C.duckdb_result
+	res apiResult
 	// chunk holds the currently active data chunk.
 	chunk DataChunk
+	// closeChunk is true after the first iteration of Next.
+	closeChunk bool
```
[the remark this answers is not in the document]
taniabogatsch: I investigated memory leaks on this PR since the bindings package exposed these allocation counters. There were a few cases where we did not close the result (in the tests), and it was a bit easier to debug/track this by tracking if a chunk has been closed. I saw that we use a flag that has to be `true` in other places, but I think we could also look at the `Ptr` of the `apiChunk`... What do you think?

scalarUDF_test.go
```diff
@@ -169,17 +169,20 @@ func (*errExecSUDF) Executor() ScalarFuncExecutor {
...
+
+	conn := openConnWrapper(t, db, context.Background())
+	defer closeConnWrapper(t, conn)
 
+	var err error
 	currentInfo, err = NewTypeInfo(TYPE_INTEGER)
 	require.NoError(t, err)
 
 	var udf *simpleSUDF
-	err = RegisterScalarUDF(c, "my_sum", udf)
```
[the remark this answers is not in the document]
taniabogatsch: Yes, true. I found a few leaks in the code (and more in the tests because we didn't close everything 😅 ), so while I worked through them, I added this. But I agree, so I'll remove these changes from this PR and keep the fixes.

rows.go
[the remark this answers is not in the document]
JAicewizard: I personally don't think it is our responsibility to track the allocations, and don't see much use (unless we expose this data in an API, but that would also create a much bigger commitment and no way back).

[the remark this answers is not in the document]
JAicewizard: (But I don't have a very strong opinion as long as it is not exposed either)

JAicewizard:
> The mappings are not a separate package; they're just a few files with build constraints in the main go-duckdb package. Should they also be moved to a separate package that lives within this package in a subdirectory?

Honestly I cannot really judge RN, maybe a subdirectory would be nice to not clutter the root? But I cannot look at them from the github UI ATM so maybe I will check back later once I can view more than 3 files

rows.go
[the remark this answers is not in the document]
taniabogatsch: True, yes - I was a bit too excited about tracking them haha - I'd like to keep the `bool`, but make sure we don't expose any other changes related to allocation verification.

appender_test.go
```diff
@@ -728,6 +728,37 @@ func TestAppenderDecimal(t *testing.T) {
...
 
+func TestAppenderStrings(t *testing.T) {
+	c, db, conn, a := prepareAppender(t, `
+	CREATE TABLE test (str VARCHAR)`)
+	defer cleanupAppender(t, c, db, conn, a)
+
+	expected := []string{
+		"I am not an inlined string no no",
+		"I am",
+		"Who wants to be inlined anyways?",
```
marcboeker, line 739: 🤣🤣🤣

arrow-mapping/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/marcboeker/go-duckdb/arrow-mapping
+
+go 1.23.4
```
marcboeker, line 3: Should we go with 1.24?

connection.go
```diff
@@ -122,83 +118,77 @@ func (c *Conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
...
 
-	if _, err := c.ExecContext(ctx, `BEGIN TRANSACTION`, nil); err != nil {
+	if _, err := conn.ExecContext(ctx, `BEGIN TRANSACTION`, nil); err != nil {
 		return nil, err
 	}
 
-	c.tx = true
-	return &tx{c}, nil
+	conn.tx = true
+	return &tx{conn}, nil
```
marcboeker, line 126: nit: Newline before return (there are multiple occurrences).

```suggestion
	conn.tx = true

	return &tx{conn}, nil
```

connection.go
```diff
@@ -122,83 +118,77 @@ func (c *Conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
...
 
+	count := m.ExtractStatements(conn.conn, query, &extractedStmts)
 	if count == 0 {
-		errMsg := C.GoString(C.duckdb_extract_statements_error(stmts))
-		C.duckdb_destroy_extracted(&stmts)
+		errMsg := m.ExtractStatementsError(extractedStmts)
+		m.DestroyExtracted(&extractedStmts)
 		if errMsg != "" {
-			return nil, 0, getDuckDBError(errMsg)
+			return extractedStmts, 0, getDuckDBError(errMsg)
```
marcboeker, line 148: nit: Even as `extractedStmts` is nil in the error case, I would return nil to be explicit. The same would apply for line 150.

go.mod
```diff
@@ -1,25 +1,31 @@
 module github.com/marcboeker/go-duckdb
 
-go 1.23
+go 1.23.4
```
marcboeker, line 3: Maybe switch to 1.24?

go.mod
```diff
@@ -30,3 +36,8 @@ require (
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace (
+	github.com/marcboeker/go-duckdb/arrow-mapping => ./arrow-mapping
```
marcboeker, line 41: TODO: Remove before merging.

go.work
```diff
@@ -0,0 +1,7 @@
+go 1.23.4
+
+use (
+	.
+	arrow-mapping
```
marcboeker, line 5: Please rename the directory to `arrow_mapping`

mapping/mapping.go
```diff
@@ -0,0 +1,355 @@
+//go:build duckdb_use_lib || duckdb_use_static_lib
+
+package mapping
+
+import (
+	bindings "github.com/duckdb/duckdb-go-bindings"
+)
+
+// ------------------------------------------------------------------ //
```
marcboeker, line 9: nit: Normal Go comments please `// Enums`

scalarUDF.go
```diff
@@ -1,12 +1,11 @@
 package duckdb
```
marcboeker, line 1: please rename to `scalar_udf.go`

statement.go
```diff
@@ -1,56 +1,52 @@
...
 	"errors"
 	"fmt"
 	"math/big"
-	"unsafe"
+
+	m "github.com/marcboeker/go-duckdb/mapping"
 )
 
-type StmtType C.duckdb_statement_type
+type StmtType m.StatementType
```
marcboeker, line 13: nit: I would call it `StatementType`

types.go
```diff
@@ -98,13 +95,13 @@ func hugeIntFromNative(i *big.Int) (C.duckdb_hugeint, error) {
...
 	if !q.IsInt64() {
-		return C.duckdb_hugeint{}, fmt.Errorf("big.Int(%s) is too big for HUGEINT", i.String())
+		return m.HugeInt{}, fmt.Errorf("big.Int(%s) is too big for HUGEINT", i.String())
 	}
 
-	return C.duckdb_hugeint{
-		lower: C.uint64_t(r.Uint64()),
-		upper: C.int64_t(q.Int64()),
-	}, nil
+	var hugeInt m.HugeInt
```
marcboeker, line 101: Maybe initialize via `NewHugeInt(r.Uint64(), q.Int64())`?

marcboeker (commented):
Thanks, @taniabogatsch, for your hard work. I've left some comments after the first overflight. Maybe we can merge it into a branch called `v2-preview` or something similar and then continue working on it, as for now, GitHub is not able to properly show the PR as it seems too large.

statement.go
taniabogatsch: We export this type, so doing that would break backward compatibility. I could add a type alias. Also, we use `Stmt` in other places, so this name is more consistent with our `Stmt` type. Personally, I'd keep it as-is, but your call! :)

taniabogatsch:
Hi Marc! Thanks for the detailed review! I am working on implementing your suggestions. I'll move some things into separate PRs and merge them (like the Go version and the mappings) so we can further decrease the scope of this one.

mapping/mapping.go
taniabogatsch: Implemented here: https://github.com/marcboeker/go-duckdb/pull/378

go.work
taniabogatsch: I went with `arrowmapping` to stay consistent with the module name (package import name). Does that also work?
See https://github.com/marcboeker/go-duckdb/pull/378

arrow_mapping.go
```diff
@@ -0,0 +1,32 @@
...
+
+package duckdb
+
+import (
+	bindings "github.com/duckdb/duckdb-go-bindings"
+)
+
+// ------------------------------------------------------------------ //
+// Pointers
+// ------------------------------------------------------------------ //
```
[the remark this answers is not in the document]
taniabogatsch: Done: https://github.com/marcboeker/go-duckdb/pull/378

README.md
```diff
@@ -68,68 +70,80 @@ That way, you can perform initialization steps in a callback function before ope
 Here's an example that configures some parameters when opening a database with `sql.OpenDB(connector)`.
 
 ```go
-connector, err := duckdb.NewConnector("/path/to/foo.db?access_mode=read_only&threads=4", func(execer driver.ExecerContext) error {
+c, err := duckdb.NewConnector(t, `/path/to/foo.db?access_mode=read_only&threads=4`, func(execer driver.ExecerContext) error {
```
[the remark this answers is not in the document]
taniabogatsch: Ah, that's likely because I turned this into a multi-line string at some point and then went back again. I usually do that for SQL strings for readability. I can revert it here!

.github/workflows/tests.yaml
```diff
@@ -6,32 +6,32 @@ on:
...
 
   test_examples:
-    name: Test examples
-    runs-on: ubuntu-latest
+    name: Test Examples
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        go: [ "1.23" ]
```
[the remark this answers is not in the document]
taniabogatsch: Done: https://github.com/marcboeker/go-duckdb/pull/379. Will merge these changes when I merge `main` into this PR again.

go.mod
taniabogatsch: Done: https://github.com/marcboeker/go-duckdb/pull/379.

scalarUDF.go
taniabogatsch: Done https://github.com/marcboeker/go-duckdb/pull/380

taniabogatsch (commented):
@marcboeker @JAicewizard, I can review this now. It loads slowly, but all the files and changes are there. Does it work better for you, now, too? I tried Brave and Safari.

README.md
```diff
@@ -1,20 +1,22 @@
...
+
 ![Tests status](https://github.com/marcboeker/go-duckdb/actions/workflows/tests.yaml/badge.svg)
 [![GoDoc](https://godoc.org/github.com/marcboeker/go-duckdb?status.svg)](https://pkg.go.dev/github.com/marcboeker/go-duckdb)
 
 ```diff
-! DuckDB just (Feb 5th, 2025) released its latest version (1.2.0).
-! go-duckdb is working on moving to that version within the next few days.
-! However, we expect some breaking changes around pre-built static libraries and the arrow dependency.
-! Therefore, it is likely that go-duckdb moves to v2 to support DuckDB v1.2.0.
+! Starting with v2.0.0, go-duckdb supports DuckDB v1.2.0.
```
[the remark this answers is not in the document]
JAicewizard: We still need to revert my first appended PR, since the second iteration makes breaking changes

scalar_udf.go
```diff
@@ -1,12 +1,11 @@
...
-#include <duckdb.h>
+void scalar_udf_callback(void *, void *, void *);
+typedef void (*scalar_udf_callback_t)(void *, void *, void *);
 
-void scalar_udf_callback(duckdb_function_info, duckdb_data_chunk, duckdb_vector);
-void udf_delete_callback(void *);
-
-typedef void (*scalar_udf_callback_t)(duckdb_function_info, duckdb_data_chunk, duckdb_vector);
+void scalar_udf_delete_callback(void *);
+typedef void (*scalar_udf_delete_callback_t)(void *);
```
JAicewizard, line 8: Really ugly to just use `void *`, moving this to mapping would also be ugly I suppose. Just keep it this way?

vector_setters.go
```diff
@@ -88,93 +84,105 @@ func setBool[S any](vec *vector, rowIdx C.idx_t, val S) error {
...
+func setHugeint[S any](vec *vector, rowIdx mapping.IdxT, val S) error {
 	var err error
-	var fv C.duckdb_hugeint
+	var fv mapping.HugeInt
 	switch v := any(val).(type) {
 	case uint8:
-		fv = C.duckdb_hugeint{lower: C.uint64_t(v)}
+		mapping.HugeIntSetLower(&fv, uint64(v))
+		mapping.HugeIntSetUpper(&fv, 0)
 	case int8:
```
JAicewizard, line 147: Did you check any benchmarks for this? This is 2 function calls for such a minor thing might be a lot. Go does not inline functions across modules IIRC. Would it be possible to just expose a type like it was before? If you don't feel like doing it now thats ok, can be looked into later after this is merged

JAicewizard (commented):
Mostly small nits, looks really good!

vector_setters.go
james-lawrence: I mean why not just set upper/lower in the same call? `mapping.HugeIntSet(&fv, uint64(v), 0)`

types.go
taniabogatsch: This would align with @james-lawrence's suggestion. Alternatively, we *could* expose the fields of the struct type directly. Wdyt?

README.md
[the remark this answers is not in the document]
taniabogatsch: Which PR is that? I remember merging this one recently. https://github.com/marcboeker/go-duckdb/pull/351

appender.go
```diff
@@ -1,84 +1,67 @@
...
-#include <duckdb.h>
-*/
-import "C"
-
 import (
 	"database/sql/driver"
 	"errors"
-	"unsafe"
+
+	"github.com/marcboeker/go-duckdb/mapping"
```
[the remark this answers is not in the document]
taniabogatsch: I am pinging @marcboeker about this. I don't have a strong opinion about the name, so I would choose whatever you (two) prefer. :)

scalar_udf.go
taniabogatsch: Yeah, I also don't like it... But I don't see a way around it. I prefer to keep the mappings what they are, i.e., just aliases, and not move any logic there.

appender.go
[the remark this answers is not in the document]
marcboeker: What about `api` as it was before, but now in its own package instead of the `duckdb` package? Or `mapper`? @JAicewizard what confuses you about `mapping`? That it is too generic?

marcboeker:
@taniabogatsch Maybe we can change the module to include to a `v2`, merge it into a `v2.0.0` branch and iterate on this one. Then we have a v2 which could be used by people and we can gather more feedback.
I would add a note, that the v2 is still experimental and its API subject to change. But I think we need to get this on the road.

taniabogatsch:
@marcboeker, that sounds good. I'll finish this up, merge it into the `v2-preview` branch, and push a tag. I'll also move the remaining comments into an issue so we can merge them separately.

Nit: Could you clean up the branches (I don't have permission to delete a branch, not even my own, haha)? There are quite a few stale ones, which might be confusing. 
- These can all go, I think: `pre-release`, `download-libs`, `windows`, `refactor-decimal-tests`
- And maybe also the dependabot things, but I'm not sure what they do.

taniabogatsch:
The changes in this PR can now be used/tried out via `_ "github.com/marcboeker/go-duckdb/v2"`. 🥳

taniabogatsch:
I also added a small note to the `README.md` on the main branch.
Screenshot 2025-03-04 at 09 31 26

go.work
JAicewizard: I think that arrowmapping matches the go idioms best as well (https://go.dev/blog/package-names)

types.go
JAicewizard: I think exposing the fields would be ideal, no harm in that right? What is the difference if we have getters and setters for both anyways

appender.go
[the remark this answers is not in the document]
JAicewizard: yeah mapping is too generic, `arrowmapping` is an example that better IMO since it clearly states that it is for arrow.

JAicewizard:
Nice work! Must have been a pain to work on this

james-lawrence:
this is wonderful news! =) looking forward to updating!

types.go
taniabogatsch: Tbh I am not even sure how to expose the fields if we alias the struct itself.
If that's possible, then I am happy to do it.

Currently, the bindings package looks like this:
```Go
type (
	Date              = C.duckdb_date
)
```
And the C header, for example, for `duckdb_date`:
```c
typedef struct {
	int32_t days;
} duckdb_date;
```

PR #386, opened 2025-03-11
Make the V2 preview branch the official version

taniabogatsch:
We've had this preview branch for a week now, and so far, there haven't been any issues (of which I am aware).

I would like this refactor / new version to go into the `main` branch.
- duckdb moved to v1.2.1, and I'd like to publish a stable (official) go-duckdb release for v1.2.0 soon and then move to v1.2.1.
- there are plenty of new C API functions/features that I'd like to implement for go-duckdb.
   - we've also other cool PRs on hold at the moment - we should pick up development on features/reviews again soon. 

We haven't decided on a different name for `mapping` yet, but changing that won't cause a breaking change, so that should not be problematic to address in the future.

What do you think @marcboeker @JAicewizard @james-lawrence?


james-lawrence:
I've been waiting for it to mainline to begin work on using it, Im a little swamped here on the home front I likely wont have spare cycles until early summer. I think its perfectly fine to mainline long as you guys are comfortable with it. =)

Issue #367, opened 2025-02-15
RegisterTableUDF panic when "SELECT count()"


vimt:
For example, 

https://github.com/marcboeker/go-duckdb/blob/686cc28f4e4005ea2c187f16cd832461cea604e8/examples/table_udf/main.go#L66

Modify the udf example query to `SELECT count() FROM increment(100)`

Run reuslt: `panic: runtime error: index out of range [-1]`

PR #391, opened 2025-03-17
[Fix] Skip row id columns in projection pushdown

taniabogatsch:
Addresses https://github.com/marcboeker/go-duckdb/issues/367.

See https://github.com/duckdb/duckdb/pull/16248.
This should be appropriately addressed post duckdb 1.3.0.


taniabogatsch:
`IsRowIdColumn` is now called `IsVirtualColumn`, and is not yet available through the C API.

PR #398, opened 2025-03-20
Use the instance cache when creating a new database

taniabogatsch:
Related issue: https://github.com/duckdb/duckdb/issues/10002
This has only been raised as an issue in the Node client, yet the same behavior can happen in other clients not using the instance cache, too. 

Any review / quick look at this solution, especially w.r.t. keeping a global variable + initializing it, would be appreciated.
In the duckdb Python package, the cache also lives in a global variable, and, afaik that's the way to solve this.


duckdb.go
```diff
@@ -15,8 +15,11 @@ import (
 	"github.com/marcboeker/go-duckdb/mapping"
 )
 
+var instanceCache mapping.InstanceCache
+
 func init() {
 	sql.Register("duckdb", Driver{})
+	instanceCache = mapping.CreateInstanceCache()
```
ayuhito, line 22: It might be better to use `sync.Once` to lazily initialise the instance cache only when `NewConnector` is called rather than calling it in `init()`.

taniabogatsch: That's a great suggestion, thanks! I will change it.

Issue #388, opened 2025-03-17
Can't scan `JSON` into `[]byte` in newer go-duckdb versions


k-anshul:
Hey 

Starting version `1.8.4` it is not possible to scan JSON into `[]byte`. The following code works in 1.8.3 but throws error in later versions : `panic: sql: Scan error on column index 0, name "json_serialize_sql('SELECT 1')": unsupported Scan, storing driver.Value type map[string]interface {} into type *[]uint8` 

IMO it is acceptable to scan JSON into []byte. Other drivers allow this.

```
db, err := sql.Open("duckdb", "")
if err != nil {
	panic(err)
}
defer db.Close()

var res []byte
err = db.QueryRowContext(context.Background(), "SELECT json_serialize_sql('SELECT 1')").Scan(&res)
if err != nil {
	panic(err)
}

fmt.Println(string(res))
```

ylhan:
Running into the same issue. Did you figure out solution?

taniabogatsch:
I'll try to have a look this week.

taniabogatsch:
Hi @k-anshul - I dove into this and dug up the PR changing this: https://github.com/marcboeker/go-duckdb/pull/325; I should've documented this change more explicitly! I apologize for that, especially given Go's policy on backward compatibility between versions. :/


IIRC, there was a confusing inconsistency between go-duckdb and duckdb's CLI - 

```sql
CREATE TABLE t1 (c1 STRUCT(index INTEGER));
INSERT INTO t1 VALUES({index: 1}), ({index: 2}), ({index: 2}), ({index: 3}), ({index: 3}), ({index: 3});
SELECT json_group_object(t2.status, t2.count) AS result
FROM (
   SELECT json_extract(t1.c1, '$.index') AS status, COUNT(*) AS count FROM t1 GROUP BY status) AS t2;
```
```
┌─────────────────────┐
│       result        │
│        json         │
├─────────────────────┤
│ {"1":1,"3":3,"2":2} │
└─────────────────────┘
```

But in go-duckdb, scanning this into `any` used to yield:
```
data = "{\"3\":3,\"2\":2,\"1\":1}"
```

I am unsure if it is possible to allow both - scanning into the expected JSON interface and `[]byte.` I just looked, and I don't think I can implement a way to scan back into the destination type without the `Scanner` interface and a custom wrapper type, but maybe I am missing something here.

[...]

PR #415, opened 2025-04-11
Handle JSON type in DatabaseTypeName and ScanType

taniabogatsch:
Fixes https://github.com/marcboeker/go-duckdb/issues/388.

cc @k-anshul

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -92,8 +92,7 @@ func (r *rows) ColumnTypeScanType(index int) reflect.Type {
 	alias := mapping.LogicalTypeGetAlias(logicalType)
 	switch alias {
 	case aliasJSON:
-		var i any
-		return reflect.TypeOf(i)
+		return reflect.TypeOf((*any)(nil)).Elem()
 	}
 
 	t := Type(mapping.ColumnType(&r.res, mapping.IdxT(index)))
--- a/rows.go
+++ b/rows.go
@@ -92,10 +92,8 @@ func (r *rows) ColumnTypeScanType(index int) reflect.Type {
 	alias := mapping.LogicalTypeGetAlias(logicalType)
 	switch alias {
 	case aliasJSON:
-		{
-			var i any
-			return reflect.TypeOf(i)
-		}
+		var i any
+		return reflect.TypeOf(i)
 	}
 
 	t := Type(mapping.ColumnType(&r.res, mapping.IdxT(index)))
--- a/rows.go
+++ b/rows.go
@@ -86,6 +86,18 @@ func (r *rows) Next(dst []driver.Value) error {
 
 // ColumnTypeScanType implements driver.RowsColumnTypeScanType.
 func (r *rows) ColumnTypeScanType(index int) reflect.Type {
+	logicalType := mapping.ColumnLogicalType(&r.res, mapping.IdxT(index))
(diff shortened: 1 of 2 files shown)


rows.go
```diff
@@ -86,6 +86,16 @@ func (r *rows) Next(dst []driver.Value) error {
...
 // ColumnTypeScanType implements driver.RowsColumnTypeScanType.
 func (r *rows) ColumnTypeScanType(index int) reflect.Type {
+	logicalType := mapping.ColumnLogicalType(&r.res, mapping.IdxT(index))
+	defer mapping.DestroyLogicalType(&logicalType)
+
+	alias := mapping.LogicalTypeGetAlias(logicalType)
+	switch alias {
+	case aliasJSON:
+		var i any
+		return reflect.TypeOf(i)
```
k-anshul, line 96: This will always return `nil`. 
If you want to return type of `any` then it should be something like `return reflect.TypeOf((*any)(nil)).Elem()`

k-anshul: May be instead of returning `ScanType` as `any` we should let it return the exact type on the basis of actual `ColumnType` i.e. do no explicit handling for JSON.

taniabogatsch: > May be instead of returning ScanType as any we should let it return the exact type on the basis of actual ColumnType i.e. do no explicit handling for JSON.

In that case, we'll return `reflect.TypeOf("")`, as that's duckdb's internal type for `JSON` columns. We only know the exact type of the JSON object once we scan it... 🤔

k-anshul: `reflect.TypeOf("")` will return `string` always. From the docs ` It should return the value type that can be used to scan types into` so returning `string` will be problematic for all cases.

taniabogatsch: > returning string will be problematic for all cases

Agreed.

> It should return the value type that can be used to scan types into

We don't know how the JSON object looks like prior to scanning it (and it is possible to call `ScanType` before the actual scan). So I am unsure about the best way to solve this (without just saying `any`)... I am now wondering if we should revert the changes that added `json.Unmarshal`... but that would mean another breaking change.

taniabogatsch: Maybe for now we return `any`? And if we consider moving to a `v3`, then we can re-discuss changing/breaking this (again)... Unsure.

k-anshul: Yeah I think returning `any` is okay.

rows.go
```diff
@@ -86,6 +86,15 @@ func (r *rows) Next(dst []driver.Value) error {
 
 // ColumnTypeScanType implements driver.RowsColumnTypeScanType.
 func (r *rows) ColumnTypeScanType(index int) reflect.Type {
+	logicalType := mapping.ColumnLogicalType(&r.res, mapping.IdxT(index))
+	defer mapping.DestroyLogicalType(&logicalType)
+
+	alias := mapping.LogicalTypeGetAlias(logicalType)
+	switch alias {
+	case aliasJSON:
+		return reflect.TypeOf((*any)(nil)).Elem()
```
murfffi, line 95: nit: Since Go 1.22, one can use `reflect.TypeFor[any]()` in this case - https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/reflect/type.go;l=2855

taniabogatsch: Oh! Cool, thanks for the hint!

PR #416, opened 2025-04-11
Do not use the instance cache for in-memory databases

taniabogatsch:
In a future PR, we can consider allowing the labelling of in-memory databases, similar to what the Python client does to cache them.
```
:memory:my_label:?...
```


duckdb.go
```diff
@@ -64,8 +68,17 @@ func NewConnector(dsn string, connInitFn func(execer driver.ExecerContext) error
...
 	defer mapping.DestroyConfig(&config)
 
+	var db mapping.Database
 	var errMsg string
-	state := mapping.GetOrCreateFromCache(GetInstanceCache(), getDBPath(dsn), &db, config, &errMsg)
+	var state mapping.State
+
+	if inMemory {
+		// Open an in-memory database.
+		state = mapping.OpenExt(getDBPath(dsn), &db, config, &errMsg)
```
k-anshul, line 77: Nit: Instead of `getDBPath(dsn)` it can just be an empty string ?

Issue #343, opened 2024-12-30
Use duckdb_bind_value to bind non-primitive parameters


taniabogatsch:
Use `duckdb_bind_value` for any non-standard parameter.

```Go
func (s *Stmt) bind(args []driver.NamedValue) error {
	if s.NumInput() > len(args) {
		return fmt.Errorf("incorrect argument count for command: have %d want %d", len(args), s.NumInput())
	}

	// FIXME (feature): we can't pass nested types as parameters (bind_value) yet

	// ...

		switch v := arg.Value.(type) {
		case bool:
			if rv := C.duckdb_bind_boolean(*s.stmt, C.idx_t(i+1), C.bool(v)); rv == C.DuckDBError {
				return errCouldNotBind
			}
		case int8:
			if rv := C.duckdb_bind_int8(*s.stmt, C.idx_t(i+1), C.int8_t(v)); rv == C.DuckDBError {
				return errCouldNotBind
			}
		case int16:
			if rv := C.duckdb_bind_int16(*s.stmt, C.idx_t(i+1), C.int16_t(v)); rv == C.DuckDBError {
				return errCouldNotBind
			}
		case int32:
			if rv := C.duckdb_bind_int32(*s.stmt, C.idx_t(i+1), C.int32_t(v)); rv == C.DuckDBError {
				return errCouldNotBind
			}
```

PR #423, opened 2025-04-18
Add support for binding lists, arrays, and structs to prepared queries

wmTJc9IK0Q:
Closes https://github.com/marcboeker/go-duckdb/issues/343

I have added support for binding list, array, and struct types to prepared queries. 

I have not yet added support for enums (should be achievable, but can be a separate PR). Maps are not possible until https://github.com/duckdb/duckdb/pull/14613 is merged upstream, which seems to be stalled right now.

Open to any feedback to get this merged, thanks.


statement.go
```diff
@@ -114,6 +115,23 @@ func (s *Stmt) ParamType(n int) (Type, error) {
 	return Type(t), nil
 }
 
+func (s *Stmt) ParamLogicalType(n int) (mapping.LogicalType, error) {
```
taniabogatsch, line 118: Let's not expose this.

statement.go
```diff
@@ -202,8 +242,21 @@ func (s *Stmt) bindComplexValue(val driver.NamedValue, n int) (mapping.State, er
...
-	case TYPE_TIMESTAMP_S, TYPE_TIMESTAMP_MS, TYPE_TIMESTAMP_NS, TYPE_LIST, TYPE_STRUCT, TYPE_MAP,
-		TYPE_ARRAY, TYPE_ENUM:
+	case TYPE_ARRAY, TYPE_LIST, TYPE_STRUCT:
+		lt, err := s.ParamLogicalType(n + 1)
+		defer mapping.DestroyLogicalType(&lt)
+		if err != nil {
+			return mapping.StateError, err
+		}
+
+		mappedVal, err := createValue(lt, val.Value)
```
taniabogatsch, line 252: `defer mapping.DestroyValue(mappedVal)`

statement.go
```diff
@@ -202,8 +242,21 @@ func (s *Stmt) bindComplexValue(val driver.NamedValue, n int) (mapping.State, er
...
+		if err != nil {
+			return mapping.StateError, addIndexToError(err, n+1)
+		}
+
+		state := mapping.BindValue(*s.preparedStmt, mapping.IdxT(n+1), *mappedVal)
+		return state, nil
```
taniabogatsch, line 258: Let's move this into its own function, like we do for the other cases.

value.go
```diff
@@ -56,3 +58,54 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+		vv, err = mapping.CreateInt8(v.(int8)), nil
+	case TYPE_SMALLINT:
+		vv, err = mapping.CreateInt16(v.(int16)), nil
+	case TYPE_INTEGER:
+		// TODO: do all int types need this casting?
+		if i, ok := v.(int); ok {
+			vv, err = mapping.CreateInt32(int32(i)), nil
+		} else {
+			vv, err = mapping.CreateInt32(v.(int32)), nil
+		}
```
taniabogatsch, line 79: We should be able to reduce this to `vv, err = mapping.CreateInt32(v.(int32)), nil`.

types.go
taniabogatsch, line 1: Can we please move `getMappedSliceValue` (`getMappedArrayValue`, `getMappedListValue`), and `getMappedStructValue` to `value.go`? :)

types.go
```diff
@@ -258,3 +258,76 @@ func getTimeTicks[T any](val T) (int64, error) {
 	base := time.Date(1970, time.January, 1, ti.Hour(), ti.Minute(), ti.Second(), ti.Nanosecond(), time.UTC)
 	return base.UnixMicro(), err
 }
+
+func getMappedArrayValue[T any](lt mapping.LogicalType, val T) (*mapping.Value, error) {
```
taniabogatsch, line 262: Let's unify this function with `getMappedListValue` and maybe call it `getMappedSliceValue`. They only differ in a single line, which we could solve like in the snippet below. We could also pass `t` (`Type`) as a parameter.

```Go
	var v mapping.Value
	t := Type(mapping.GetTypeId(lt))
	if t == TYPE_ARRAY {
		v = mapping.CreateArrayValue(childType, childValues)
	} else {
		v = mapping.CreateListValue(childType, childValues)
	}
	
	return &v, nil
```

types.go
```diff
@@ -258,3 +258,76 @@ func getTimeTicks[T any](val T) (int64, error) {
...
+	childType := mapping.ArrayTypeChildType(lt)
+	defer mapping.DestroyLogicalType(&childType)
+
+	vSlice, err := toAnySlice(val)
+	if err != nil {
+		return nil, fmt.Errorf("could not cast %T to []any: %s", val, err)
+	}
+
+	for _, v := range vSlice {
+		vv, err := createValue(childType, v)
```
taniabogatsch, line 273: `defer mapping.DestroyValue(vv)`

statement.go
```diff
@@ -186,6 +204,28 @@ func (s *Stmt) bindTime(val driver.NamedValue, t Type, n int) (mapping.State, er
 	return state, nil
 }
 
+func toAnySlice(v any) ([]any, error) {
```
taniabogatsch, line 207: Maybe we can unify (part of) this function with the `func extractSlice[S any](vec *vector, val S) ([]any, error)` function in `vector_setters.go`? And move it to that file, or to `value.go`.

statement.go
```diff
@@ -186,6 +204,28 @@ func (s *Stmt) bindTime(val driver.NamedValue, t Type, n int) (mapping.State, er
...
+	if rv.Kind() == reflect.Slice || rv.Kind() == reflect.Array {
+		result := make([]any, rv.Len())
+		for i := 0; i < rv.Len(); i++ {
+			result[i] = rv.Index(i).Interface()
+		}
+		return result, nil
+	}
+
+	// Handle single values by wrapping in a slice
+	return []any{v}, nil
```
taniabogatsch, line 226: I prefer if we return an invalid input error here, as strictly speaking the user did not provide an `ARRAY` or `LIST` as a parameter.

types.go
```diff
@@ -258,3 +258,76 @@ func getTimeTicks[T any](val T) (int64, error) {
...
+
+	childCount := mapping.StructTypeChildCount(lt)
+
+	for i := mapping.IdxT(0); i < childCount; i++ {
+		childName := mapping.StructTypeChildName(lt, i)
+		childType := mapping.StructTypeChildType(lt, i)
+		defer mapping.DestroyLogicalType(&childType)
+		v, exists := vMap[childName]
+		if exists {
+			vv, err := createValue(childType, v)
```
taniabogatsch, line 321: `defer mapping.DestroyValue(vv)`

types.go
```diff
@@ -258,3 +258,76 @@ func getTimeTicks[T any](val T) (int64, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("could not create value %s", err)
+			}
+			values = append(values, *vv)
+		} else {
+			values = append(values, mapping.CreateNullValue())
```
taniabogatsch, line 327:
```Go
			nullVal := mapping.CreateNullValue()
			defer mapping.DestroyValue(&nullVal)
			values = append(values, nullVal)
```

wmTJc9IK0Q (commented):
@taniabogatsch Thanks for the review. I have addressed most of the feedback. We just need to decide on the path forward on the int casts in value.go

statement.go
```diff
@@ -187,6 +204,24 @@ func (s *Stmt) bindTime(val driver.NamedValue, t Type, n int) (mapping.State, er
...
+// Used for binding Array, List, Struct. In the future, also Map and Union
+func (s *Stmt) bindCompositeValue(val driver.NamedValue, n int) (mapping.State, error) {
+	lt, err := s.paramLogicalType(n + 1)
+	defer mapping.DestroyLogicalType(&lt)
+	if err != nil {
+		return mapping.StateError, err
+	}
+
+	mappedVal, err := createValue(lt, val.Value)
+	// defer mapping.DestroyValue(mappedVal)
```
taniabogatsch: Can we please uncomment this?

I dove into the C/C++ side, to ensure that we don't leak memory by not destroying this value. We move the value in `BindValue`- so strictly speaking, we do not need to destroy here, however, for correct/consistent API usage, I am still in favour of calling `mapping.DestroyValue` on the `mapping.Value` wrapper - this will make it future proof, in case the wrapper contains more than the moved-out value in the future.

```cpp
duckdb_state duckdb_bind_value(duckdb_prepared_statement prepared_statement, idx_t param_idx, duckdb_value val) {
 // ...
	wrapper->values[identifier] = duckdb::BoundParameterData(*value);
	return DuckDBSuccess;
}
```
`move` in the constructor.
```cpp
	explicit BoundParameterData(Value val) : value(std::move(val)), return_type(GetDefaultType(value.type())) {
	}
```

So we move the value into the prepared statement wrapper.
```cpp
struct PreparedStatementWrapper {
	//! Map of name -> values
	case_insensitive_map_t<BoundParameterData> values;
	unique_ptr<PreparedStatement> statement;
};
```

value.go
```diff
@@ -1,6 +1,10 @@
 package duckdb
 
 import (
+	"fmt"
+	"reflect"
+
+	duckdb_go_bindings "github.com/duckdb/duckdb-go-bindings/darwin-arm64"
```
taniabogatsch: Remove this import and all usages - we must always go through the `mapping` package.

value.go
taniabogatsch: That's the expected error, if the Go type does not match the expected bind type.

Adjusting the test like so: `map[string]any{"v": "baz", "i": int32(42)},` (adding the cast) solves this.
We get a similar failure when, for example, passing a boolean instead of a string: `map[string]any{"v": false, "i": int32(42)},`.

We might want to properly solve this in the future by not just assuming that the cast will succeed everywhere (`mapping.CreateBool(v.(bool))`, ...). Instead, we could use `i, ok := v.(type)` to throw better error messages for all cases...

taniabogatsch (commented):
Thanks for implementing the review feedback! I just had another pass and only found a few minor things, so I quickly added and pushed them myself.

I've also left a few comments to explain what I did.

What do you think, does this look ready from your side now? :)

wmTJc9IK0Q:
@taniabogatsch Thanks for the small fixes. This looks good to me.

wmTJc9IK0Q:
I fixed one bug where you could not bind an `[]any` parameter. This is especially important since the library doesn't currently support scanning into a specific slice type. So if you want to use the same type as an argument to a query and to scan into, you need to use `[]any`

taniabogatsch (approved):

taniabogatsch:
Thank you! And great catch with the `[]any` parameter! :)

PR #427, opened 2025-04-24
fix: switch Decimal methods to value receivers

### LINKED ISSUE #424: Decimal methods aren't exposed on scanned values
With latest `go-duckdb` , the output of this program:

```go
check(err)
defer db.Close()
var result any
// SQL statement adapted from https://duckdb.org/docs/stable/sql/data_types/numeric.html#fixed-point-decimals
err = db.QueryRow("SELECT 1.3-0.7").Scan(&result)
check(err)
fmt.Printf("%v %T", result, result)
```

is 

```
{3 1 6} duckdb.Decimal
```

while the more useful output of `fmt.Printf("%v", result)` would be `0.6` .

Looking it at the [source](https://github.com/marcboeker/go-duckdb/blob/main/types.go#L148), Decimal does attempt to implement `fmt.Stringer` with a String() method but on a pointer receiver. If we replace, `var result any` with `var result *duckdb.Decimal`, we get the expected output but passing pointer to a pointer to `Scan()` is not obvious and counter to the typical use of `database/sql` API . Using `var result duckdb.Decimal` doesn't give the right result either and prints the same as with `var result any` .

I saw that `Decimal.String()` was added relatively recently in #308 and there are even unit tests. The tests work like this:

```go
r := db.QueryRow(fmt.Sprintf("SELECT %s", test.input))
var fs Decimal
require.NoError(t, r.Scan(&fs))
require.Equal(t, test.want, fs.String())
```

This works almost by chance - because in `fs.String()` Go auto-magically calls the pointer receiver method even though `fs` is not a pointer.

If we replace, `fs.String()` above with `fmt.Sprint(fs)` which should be equivalent, the test fails. Additionally, if we replace `fs.String()` with `fs.(fmt.Stringer).String()` in the test, the code doesn't compile.

I propose that the methods of Decimal are implemented on a value receiver. Efficiency cost will be minimal, if any, since the main field `Decimal.Value` is already a pointer. If that makes sense, I can send a PR.

murfffi:
Switches the Decimal methods from pointer to value receivers.

The ScanType of DECIMAL columns is Decimal, not *Decimal. Users would expect that methods are available on the exact type returned from ScanType, and this fix implements it. For example, this fix allows Decimal to be used where fmt.Stringer is expected, e.g. in the `fmt.*Print*` functions.

Closes #424

### DIFF
--- a/types.go
+++ b/types.go
@@ -151,7 +151,7 @@ type Decimal struct {
 	Value *big.Int
 }
 
-func (d *Decimal) Float64() float64 {
+func (d Decimal) Float64() float64 {
 	scale := big.NewInt(int64(d.Scale))
 	factor := new(big.Float).SetInt(new(big.Int).Exp(big.NewInt(10), scale, nil))
 	value := new(big.Float).SetInt(d.Value)
@@ -160,7 +160,7 @@ func (d *Decimal) Float64() float64 {
 	return f
 }
 
-func (d *Decimal) String() string {
+func (d Decimal) String() string {
 	// Get the sign, and return early, if zero.
 	if d.Value.Sign() == 0 {
 		return "0"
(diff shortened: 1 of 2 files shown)


taniabogatsch:
Thanks!

PR #432, opened 2025-05-03
Fix validity corruption bug when appending data chunks

### LINKED ISSUE #431: Union fails to append in various scenarios
# Background

I started investigating a bug I am having with using a column of the type `MAP(VARCHAR, UNION(i INT, s VARCHAR))`. In my application I am using the Appender API to insert these rows. Rows can have multiple keys with mixed value types, hence the use of the UNION.

In my application's case, I am not having any issues with appending data, however when reading data I get problems where duckdb fails to extract a key from the map unless I also specify the tag name, which is not known. For example, with a schema like:

```sql
CREATE TABLE test(id INT, map_col MAP(VARCHAR, UNION(i INT, s VARCHAR))
```

A query like this will fail with a segmentation fault when queried from go-duckdb or an internal error `INTERNAL Error: Attempted to access index 147 within vector of size 4` when using the duckdb CLI:

```sql
SELECT map_col['key1'] FROM test;
```

But this query will work fine:
```sql
SELECT map_col['key1'].s FROM test;
```

As far as I can tell, this problem seems to be specific to maps that contain VARCHAR and INT in the same map. In those cases, selecting just a map key that has a VARCHAR value fails, however selecting the INT key does not.

# Reproduction

So I went to implement a test to reproduce this issue in go-duckdb, but here I am getting problems even just appending the rows. Now I am getting a SIGBUS if I insert a map value that has both an INT and VARCHAR. I have posted my repro as a PR here: https://github.com/marcboeker/go-duckdb/pull/430. Note that sometimes the append just hangs, but most of the time it fails with this SIGBUS error:

```
=== RUN   TestAppenderNested
SIGSEGV: segmentation violation
PC=0x103313290 m=8 sigcode=2 addr=0x0
signal arrived during cgo execution

goroutine 7 gp=0x1400014a700 m=8 mp=0x14000580008 [syscall]:
runtime.cgocall(0x102c0f91c, 0x140001492d8)
        /opt/homebrew/Cellar/go/1.24.1/libexec/src/runtime/cgocall.go:167 +0x44 fp=0x140001492a0 sp=0x14000149260 pc=0x10299f724
github.com/duckdb/duckdb-go-bindings/darwin-arm64._Cfunc_duckdb_vector_assign_string_element(0x10c428540, 0x36d, 0x11ce0f200)
        _cgo_gotypes.go:4767 +0x30 fp=0x140001492d0 sp=0x140001492a0 pc=0x102a87110
github.com/duckdb/duckdb-go-bindings/darwin-arm64.VectorAssignStringElement.func1(...)
        /Users/wmTJc9IK0Q/go/pkg/mod/github.com/duckdb/duckdb-go-bindings/darwin-arm64@v0.1.9/bindings.go:2532
github.com/duckdb/duckdb-go-bindings/darwin-arm64.VectorAssignStringElement({0x10c428540}, 0x36d, {0x1042f2d5f?, 0x0?})
        /Users/wmTJc9IK0Q/go/pkg/mod/github.com/duckdb/duckdb-go-bindings/darwin-arm64@v0.1.9/bindings.go:2532 +0x9c fp=0x14000149330 sp=0x140001492d0 pc=0x102a8d68c
github.com/marcboeker/go-duckdb/v2.setBytes[...](0x1?, 0x1043e2ce8?, {0x104c59760?, 0x140000e7800?})
        /Users/wmTJc9IK0Q/src/go-duckdb/vector_setters.go:200 +0x70 fp=0x140001493d0 sp=0x14000149330 pc=0x102bdf2a0
github.com/marcboeker/go-duckdb/v2.(*vector).init.(*vector).initBytes.func36(0x104c76720?, 0x14000149748?, {0x104c59760?, 0x140000e7800?})
        /Users/wmTJc9IK0Q/src/go-duckdb/vector.go:271 +0x40 fp=0x14000149410 sp=0x140001493d0 pc=0x102b601e0
github.com/marcboeker/go-duckdb/v2.setStruct[...](0x140001013c0?, 0x36d, {0x104c8b8c0, 0x140000e7800?})
        /Users/wmTJc9IK0Q/src/go-duckdb/vector_setters.go:313 +0x5e8 fp=0x14000149780 sp=0x14000149410 pc=0x102bdd438
github.com/marcboeker/go-duckdb/v2.(*vector).initStruct.func2(0x14a00000149808?, 0x14000580008?, {0x104c8b8c0?, 0x140000e7800?})
        /Users/wmTJc9IK0Q/src/go-duckdb/vector.go:422 +0x40 fp=0x140001497c0 sp=0x14000149780 pc=0x102b5f6f0
github.com/marcboeker/go-duckdb/v2.(*DataChunk).SetValue(0x1055cc108?, 0x14000149868?, 0x400149888?, {0x104c8b8c0?, 0x140000e7800?})
        /Users/wmTJc9IK0Q/src/go-duckdb/data_chunk.go:59 +0x54 fp=0x14000149830 sp=0x140001497c0 pc=0x102b590a4
github.com/marcboeker/go-duckdb/v2.(*Appender).appendRowSlice(0x1400010abe0, {0x14000149ab8, 0xe, 0x140000fc840?})
        /Users/wmTJc9IK0Q/src/go-duckdb/appender.go:163 +0x13c fp=0x140001498b0 sp=0x14000149830 pc=0x102b1e64c
github.com/marcboeker/go-duckdb/v2.(*Appender).AppendRow(0x104c8b6e0?, {0x14000149ab8?, 0x2?, 0x0?})
        /Users/wmTJc9IK0Q/src/go-duckdb/appender.go:140 +0x28 fp=0x140001498e0 sp=0x140001498b0 pc=0x102b1e478
github.com/marcboeker/go-duckdb/v2.appendNestedData[...](0x1400014a540, 0x1400010abe0, {0x14000180000, 0x0?, 0x19c?})
        /Users/wmTJc9IK0Q/src/go-duckdb/appender_test.go:1059 +0x248 fp=0x14000149bc0 sp=0x140001498e0 pc=0x102bc96f8
github.com/marcboeker/go-duckdb/v2.TestAppenderNested(0x1400014a540)
        /Users/wmTJc9IK0Q/src/go-duckdb/appender_test.go:256 +0xc8 fp=0x14000149f60 sp=0x14000149bc0 pc=0x102b30968
testing.tRunner(0x1400014a540, 0x104cc36f8)
        /opt/homebrew/Cellar/go/1.24.1/libexec/src/testing/testing.go:1792 +0xe4 fp=0x14000149fb0 sp=0x14000149f60 pc=0x102a278d4
testing.(*T).Run.gowrap1()
        /opt/homebrew/Cellar/go/1.24.1/libexec/src/testing/testing.go:1851 +0x2c fp=0x14000149fd0 sp=0x14000149fb0 pc=0x102a2875c
runtime.goexit({})
        /opt/homebrew/Cellar/go/1.24.1/libexec/src/runtime/asm_arm64.s:1223 +0x4 fp=0x14000149fd0 sp=0x14000149fd0 pc=0x1029aa5e4
created by testing.(*T).Run in goroutine 1
        /opt/homebrew/Cellar/go/1.24.1/libexec/src/testing/testing.go:1851 +0x374

... all of the parked goroutines ...

r0      0x0
r1      0x0
r2      0x11ce0f200
r3      0x199568930
r4      0xdb0
r5      0x14000148000
r6      0x1
r7      0x104ccd0c0
r8      0x0
r9      0xf
r10     0x0
r11     0x1
r12     0x0
r13     0x171206f00
r14     0x9e3779b97f4a7c55
r15     0x144400
r16     0x1995689e0
r17     0x140001493b0
r18     0x0
r19     0x11ce0f200
r20     0xf
r21     0x10c428540
r22     0x10d2a7200
r23     0x0
r24     0x0
r25     0x14000044df8
r26     0x104cc4a00
r27     0x818
r28     0x140005821c0
r29     0x171206e60
lr      0x10331324c
sp      0x171206e10
pc      0x103313290
fault   0x0
exit status 2
FAIL    github.com/marcboeker/go-duckdb/v2      0.380s
```

# Next steps

I am looking for any advice on how to debug this - I am not yet sure if this is a go-duckdb issue or a duckdb one. Clearly the data chunk is corrupted in some way as none of these problems seem to surface if I do not use the appender API.

wmTJc9IK0Q:
Fixes #431

Fixes a seg fault that was happening due to the validity map not being reloaded when new data chunks are appended to an appender. For the full explanation and investigation see the issue.

I think this issue is a bug resulting from the change in https://github.com/marcboeker/go-duckdb/pull/428 and became more apparent in usages of Union where all non-populated members (in child vectors) are set to null.

### DIFF
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -87,23 +87,22 @@ func (chunk *DataChunk) initFromTypes(types []mapping.LogicalType, writable bool
 	}
 
 	chunk.chunk = mapping.CreateDataChunk(types)
-	mapping.DataChunkSetSize(chunk.chunk, mapping.IdxT(GetDataChunkCapacity()))
-
-	// Initialize the vectors and their child vectors.
-	for i := 0; i < columnCount; i++ {
-		v := mapping.DataChunkGetVector(chunk.chunk, mapping.IdxT(i))
-		chunk.columns[i].initVectors(v, writable)
-	}
+	chunk.initVectors(writable)
 
 	return nil
 }
 
-func (chunk *DataChunk) reset() {
+func (chunk *DataChunk) reset(writable bool) {
 	mapping.DataChunkReset(chunk.chunk)
+	chunk.initVectors(writable)
+}
 
-	for i, col := range chunk.columns {
+func (chunk *DataChunk) initVectors(writable bool) {
+	mapping.DataChunkSetSize(chunk.chunk, mapping.IdxT(GetDataChunkCapacity()))
+
+	for i := 0; i < len(chunk.columns); i++ {
 		v := mapping.DataChunkGetVector(chunk.chunk, mapping.IdxT(i))
-		col.initVectors(v, true)
+		chunk.columns[i].initVectors(v, writable)
 	}
 }
 
--- a/data_chunk.go
+++ b/data_chunk.go
@@ -98,6 +98,15 @@ func (chunk *DataChunk) initFromTypes(types []mapping.LogicalType, writable bool
 	return nil
 }
 
+func (chunk *DataChunk) reset() {
+	mapping.DataChunkReset(chunk.chunk)
+
+	for i, col := range chunk.columns {
+		v := mapping.DataChunkGetVector(chunk.chunk, mapping.IdxT(i))
+		col.initVectors(v, true)
+	}
+}
+
 func (chunk *DataChunk) initFromDuckDataChunk(inputChunk mapping.DataChunk, writable bool) error {
 	columnCount := mapping.DataChunkGetColumnCount(inputChunk)
 	chunk.columns = make([]vector, columnCount)
--- a/appender.go
+++ b/appender.go
@@ -182,7 +182,7 @@ func (a *Appender) appendDataChunk() error {
 		return getDuckDBError(mapping.AppenderError(a.appender))
 	}
 
-	a.chunk.reset()
+	a.chunk.reset(true)
 	a.rowCount = 0
 
 	return nil
--- a/appender.go
+++ b/appender.go
@@ -182,7 +182,7 @@ func (a *Appender) appendDataChunk() error {
 		return getDuckDBError(mapping.AppenderError(a.appender))
 	}
 
-	mapping.DataChunkReset(a.chunk.chunk)
+	a.chunk.reset()
 	a.rowCount = 0
 
 	return nil
--- a/appender_test.go
+++ b/appender_test.go
@@ -940,9 +940,7 @@ func TestAppenderAppendDataChunk(t *testing.T) {
 		require.NoError(t, a.AppendRow(i, Union{Value: "str2", Tag: "s"}))
 		require.NoError(t, a.AppendRow(i, nil))
 	}
-
-	err := a.Flush()
-	require.NoError(t, err)
+	require.NoError(t, a.Flush())
 }
 
 func BenchmarkAppenderNested(b *testing.B) {
--- a/appender_test.go
+++ b/appender_test.go
@@ -929,6 +929,22 @@ func TestAppenderUnion(t *testing.T) {
 	require.Equal(t, len(testCases), i)
 }
 
+func TestAppenderAppendDataChunk(t *testing.T) {
+	// Ensures that appending multiple data chunks correctly resets the previous chunk.
+
+	c, db, conn, a := prepareAppender(t, `CREATE TABLE test(id INT, attr UNION(i INT, s VARCHAR))`)
+	defer cleanupAppender(t, c, db, conn, a)
+
+	// Add enough rows to overflow several chunks.
+	for i := 0; i < GetDataChunkCapacity()*3; i++ {
+		require.NoError(t, a.AppendRow(i, Union{Value: "str2", Tag: "s"}))
+		require.NoError(t, a.AppendRow(i, nil))
+	}
+
+	err := a.Flush()
+	require.NoError(t, err)
+}
+
 func BenchmarkAppenderNested(b *testing.B) {
 	c, db, conn, a := prepareAppender(b, createNestedDataTableSQL)
 	defer cleanupAppender(b, c, db, conn, a)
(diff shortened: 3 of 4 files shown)


data_chunk.go
```diff
@@ -98,6 +98,15 @@ func (chunk *DataChunk) initFromTypes(types []mapping.LogicalType, writable bool
 	return nil
 }
 
+func (chunk *DataChunk) reset() {
+	mapping.DataChunkReset(chunk.chunk)
+
+	for i, col := range chunk.columns {
+		v := mapping.DataChunkGetVector(chunk.chunk, mapping.IdxT(i))
+		col.initVectors(v, true)
```
wmTJc9IK0Q, line 106: I don't know if we should re-init all vectors, but it seems the safest way to ensure all child-child vectors of nested types have their validity map reloaded

wmTJc9IK0Q:
@taniabogatsch I also just found that inserting NULL values into a Union column is failing with a similar seg fault, but this PR doesn't yet fix this. Reverting https://github.com/marcboeker/go-duckdb/pull/428 makes both go away. I think we should do that until we have these fixed.

taniabogatsch:
> I also just found that inserting NULL values into a Union column is failing with a similar seg fault

Could you add the test for that to this PR, as well? :) Then, I can check it out and have a look myself, and, indeed, revert some changes, if we cannot fix this.

wmTJc9IK0Q:
@taniabogatsch I updated the test to append `nil` values as well. You can see the failure by running the test with: `go test -v -run "TestAppenderAppendDataChunk\$"`

data_chunk.go
taniabogatsch: Yes, I agree that we need to re-initialize the vectors!

taniabogatsch:
> found that inserting NULL values into a Union column is failing with a similar seg fault, but this PR doesn't yet fix this 

I think this PR actually fixed it, there was just a small bug in the fix.

I cannot reproduce the seg fault with the `nil` test anymore, also not in the CI test runs. [My guess is this](https://go.dev/tour/moretypes/16):

> When ranging over a slice, two values are returned for each iteration. The first is the index, and the second is a copy of the element at that index.

And specifically this change:
```diff
- for i, col := range chunk.columns {
+	for i := 0; i < len(chunk.columns); i++ {
		v := mapping.DataChunkGetVector(chunk.chunk, mapping.IdxT(i))
-		col.initVectors(v, true)
+		chunk.columns[i].initVectors(v, writable)
	}
```

So I think we copied the child vector, and then re-initialized it in the copy, instead of the actual child vector. Would that make sense? Do you think the PR is ready to be merged now? :)

wmTJc9IK0Q:
@taniabogatsch Just getting around to this - yes that change makes sense to me. Thanks.

PR #445, opened 2025-05-20
Stream arrow reader

datbth:
Merge `main` into https://github.com/marcboeker/go-duckdb/pull/317


taniabogatsch (changes requested):
Hi @datbth, sorry for the delayed review! The changes look good to me! Could we add a test that (concurrently) uses the streaming record reader?

PR #455, opened 2025-05-24
Use runtime reflection type to bind query parameters

### LINKED ISSUE #450: Complex typed values could not be passed as function parameter
```go

func TestPrepareComplexFunctionParameter(t *testing.T) {
	db := openDbWrapper(t, ``)
	defer closeDbWrapper(t, db)

	ctx := context.Background()
	createTable(t, db, `CREATE OR REPLACE TABLE arr_test(
		arr_int INTEGER[]
	)`)

	// Insert parameters.
	_, err := db.ExecContext(ctx, `INSERT INTO arr_test VALUES (?)`,
		[]int32{1, 1})
	require.NoError(t, err)

	prepared, err := db.Prepare(`SELECT  list_avg(?)`)
	defer closePreparedWrapper(t, prepared)
	require.NoError(t, err)

	var arr Composite[float32]

	// Test with `any` slice types.
	err = prepared.QueryRow([]int32{1, 1}).Scan(&arr)
	require.NoError(t, err)
}
```

Got following error:
=== RUN   TestPrepareComplexFunctionParameter
    statement_test.go:482: 
        	Error Trace:	/Users/bytedance/Code/go-duckdb/statement_test.go:482
        	Error:      	Received unexpected error:
        	            	could not bind parameter
        	            	unsupported data type: INVALID: index: 1
        	Test:       	TestPrepareComplexFunctionParameter
--- FAIL: TestPrepareComplexFunctionParameter (0.02s)

Seems the `ParamType` function could not properly resolve parameter type from query:

```go
func ParamType(preparedStmt PreparedStatement, index IdxT) Type {
	return C.duckdb_param_type(preparedStmt.data(), index)
}
```

Here are some similar case:

```sql
SELECT  idx, list_cosine_similarity(?, embedding) from  data_embeddings
```

crisismaple:
Fix #450 by replacing the `tryBindComplexValue`, support dynamic create query parameter values by runtime reflection type when DB engine is not able to determine the logic type of parameter from query statement

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -276,6 +276,10 @@ func (s *Stmt) bindComplexValue(val driver.NamedValue, n int, t Type, name strin
 	}
 
 	switch t {
+	case TYPE_UUID:
+		if ss, ok := val.Value.(fmt.Stringer); ok {
+			return mapping.BindVarchar(*s.preparedStmt, mapping.IdxT(n+1), ss.String()), nil
+		}
 	case TYPE_TIMESTAMP, TYPE_TIMESTAMP_TZ, TYPE_TIMESTAMP_S, TYPE_TIMESTAMP_MS, TYPE_TIMESTAMP_NS:
 		return s.bindTimestamp(val, t, n)
 	case TYPE_DATE:
--- a/statement.go
+++ b/statement.go
@@ -5,10 +5,8 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
-	"math/big"
-	"time"
-
 	"github.com/marcboeker/go-duckdb/mapping"
+	"math/big"
 )
 
 type StmtType mapping.StatementType
@@ -260,21 +258,21 @@ func (s *Stmt) bindCompositeValue(val driver.NamedValue, n int) (mapping.State, 
 	return state, nil
 }
 
-func (s *Stmt) tryBindComplexValue(val driver.NamedValue, t Type, n int) (mapping.State, error) {
-	switch val.Value.(type) {
-	case time.Time:
-		// Fallback to TIMESTAMP, if we cannot know the exact type.
-		return s.bindTimestamp(val, TYPE_TIMESTAMP, n)
+func (s *Stmt) bindValueByReflection(val driver.NamedValue, n int) (mapping.State, error) {
+	_, mappedVal, err := createValueByReflection(val.Value)
+	defer mapping.DestroyValue(mappedVal)
+	if err != nil {
+		return mapping.StateError, addIndexToError(err, n+1)
 	}
-	name := typeToStringMap[t]
-	return mapping.StateError, addIndexToError(unsupportedTypeError(name), n+1)
+	state := mapping.BindValue(*s.preparedStmt, mapping.IdxT(n+1), *mappedVal)
+	return state, nil
 }
 
 func (s *Stmt) bindComplexValue(val driver.NamedValue, n int, t Type, name string) (mapping.State, error) {
 	// We could not resolve this parameter when binding the query.
 	// Fall back to the Go type.
 	if t == TYPE_INVALID {
-		return s.tryBindComplexValue(val, t, n)
+		return s.bindValueByReflection(val, n)
 	}
 
 	switch t {
--- a/statement.go
+++ b/statement.go
@@ -240,7 +240,7 @@ func (s *Stmt) bindJSON(val driver.NamedValue, n int) (mapping.State, error) {
 	return mapping.StateError, addIndexToError(unsupportedTypeError("JSON interface, need []byte or string"), n+1)
 }
 
-func (s *Stmt) bindUUID(val driver.NamedValue, t Type, n int) (mapping.State, error) {
+func (s *Stmt) bindUUID(val driver.NamedValue, n int) (mapping.State, error) {
 	if ss, ok := val.Value.(fmt.Stringer); ok {
 		return mapping.BindVarchar(*s.preparedStmt, mapping.IdxT(n+1), ss.String()), nil
 	}
@@ -285,7 +285,7 @@ func (s *Stmt) bindComplexValue(val driver.NamedValue, n int, t Type, name strin
 
 	switch t {
 	case TYPE_UUID:
-		return s.bindUUID(val, t, n)
+		return s.bindUUID(val, n)
 	case TYPE_TIMESTAMP, TYPE_TIMESTAMP_TZ, TYPE_TIMESTAMP_S, TYPE_TIMESTAMP_MS, TYPE_TIMESTAMP_NS:
 		return s.bindTimestamp(val, t, n)
 	case TYPE_DATE:
--- a/statement.go
+++ b/statement.go
@@ -240,6 +240,13 @@ func (s *Stmt) bindJSON(val driver.NamedValue, n int) (mapping.State, error) {
 	return mapping.StateError, addIndexToError(unsupportedTypeError("JSON interface, need []byte or string"), n+1)
 }
 
+func (s *Stmt) bindUUID(val driver.NamedValue, t Type, n int) (mapping.State, error) {
+	if ss, ok := val.Value.(fmt.Stringer); ok {
+		return mapping.BindVarchar(*s.preparedStmt, mapping.IdxT(n+1), ss.String()), nil
+	}
+	return mapping.StateError, addIndexToError(unsupportedTypeError(unknownTypeErrMsg), n+1)
+}
+
 // Used for binding Array, List, Struct. In the future, also Map and Union
 func (s *Stmt) bindCompositeValue(val driver.NamedValue, n int) (mapping.State, error) {
 	lt, err := s.paramLogicalType(n + 1)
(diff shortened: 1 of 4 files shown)


taniabogatsch:
Hi @crisismaple - thanks for the PR! I'll leave a review soon. :)

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+	}
+
+	value := reflect.ValueOf(i)
+	kind := value.Kind()
+
+	switch kind {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:
+		return value.IsNil()
+	}
+	return false
```
taniabogatsch, line 176: Let's make `false` the default value of the `switch`.

```Go
	switch kind {
	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:
		return value.IsNil()
	default:
		return false
	}
}
```

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+		vv, err := getMappedSliceValueByReflection(r.Interface(), true)
+		return TYPE_ARRAY, vv, err
 	default:
-		return nil, unsupportedTypeError(reflect.TypeOf(v).Name())
+		return t, nil, unsupportedTypeError(reflect.TypeOf(v).Name())
 	}
+}
 
-	return &vv, err
+func getMappedSliceValueByReflection[T any](val T, isArray bool) (*mapping.Value, error) {
```
taniabogatsch, line 245: Maybe rename to `tryGetMappedSliceValue`?

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+	}
+	vSlice, err := extractSlice(val)
+	if err != nil {
+		return nil, fmt.Errorf("could not cast %T to []any: %s", val, err)
+	}
+	var childValues []mapping.Value
+	defer destroyValueSlice(childValues)
+	if len(vSlice) == 0 {
+		lt := mapping.CreateLogicalType(TYPE_SQLNULL)
+		defer destroyLogicalTypes(toPtr([]mapping.LogicalType{lt}))
```
taniabogatsch, line 258: `defer mapping.DestroyLogicalType(&lt)`

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+		if elementType == TYPE_INVALID || elementType == TYPE_SQLNULL {
+			elementType = et
+		}
+		childValues = append(childValues, *vv)
+	}
+	if elementType == TYPE_INVALID {
+		return nil, unsupportedTypeError(reflect.TypeOf(val).Name())
+	}
+	lt := mapping.CreateLogicalType(elementType)
+	defer destroyLogicalTypes(toPtr([]mapping.LogicalType{lt}))
```
taniabogatsch, line 276: `defer mapping.DestroyLogicalType(&lt)`

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+	kind := value.Kind()
+
+	switch kind {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:
+		return value.IsNil()
+	}
+	return false
+}
+
+func createValueByReflection(v any) (Type, *mapping.Value, error) {
```
taniabogatsch, line 179: We should return `typeInfo` to support nested types (lists in lists in lists, etc.). See my other comment below.

value.go
```diff
@@ -67,45 +68,213 @@ func getValue(info TypeInfo, v mapping.Value) (any, error) {
...
+		}
+		if elementType == TYPE_INVALID || elementType == TYPE_SQLNULL {
+			elementType = et
+		}
+		childValues = append(childValues, *vv)
+	}
+	if elementType == TYPE_INVALID {
+		return nil, unsupportedTypeError(reflect.TypeOf(val).Name())
+	}
+	lt := mapping.CreateLogicalType(elementType)
```
taniabogatsch, line 275: It is not sufficient to only return the child `Type` here - `mapping.CreateLogicalType` only works for primitive types. For a `LIST` containing another `LIST`, we need to use `CreateListType`, for example. `typeInfo` does that, if set correctly: `func (info *typeInfo) logicalType() mapping.LogicalType`.

statement_test.go
```diff
@@ -430,7 +430,9 @@ func TestBindWithoutResolvedParams(t *testing.T) {
 	// Type without a fallback.
```
taniabogatsch, line 430: Let's update the comment, too. Also, can we add another test for a type we do not yet support (e.g., `UNION`, I think)?

statement_test.go
```diff
@@ -534,3 +536,60 @@ func TestBindJSON(t *testing.T) {
...
+
+	var res Composite[[]any]
+	err = dynamicPrepare.QueryRow([]any{}).Scan(&res)
+	require.NoError(t, err)
+	require.Equal(t, 0, len(res.Get()))
+
+	// The statement type has already been bind as a SQL_NULL LIST in previous query
+	err = dynamicPrepare.QueryRow([]any{1}).Scan(&res)
+	require.NoError(t, err)
+	require.Equal(t, []any{nil}, res.Get())
```
taniabogatsch, line 578: Is this correct? Why are we not rebinding here?

statement_test.go
```diff
@@ -534,3 +536,60 @@ func TestBindJSON(t *testing.T) {
...
+
+	ptrPrepare, err := db.Prepare(`SELECT * from (VALUES (?))`)
+	defer closePreparedWrapper(t, ptrPrepare)
+	require.NoError(t, err)
+
+	err = ptrPrepare.QueryRow([]any{toPtr(1)}).Scan(&res)
+	require.NoError(t, err)
+	require.Equal(t, []any{int64(1)}, res.Get())
```
taniabogatsch, line 586: Can we add a test with nested pointers? To test the `getPointerValue` a bit more?

statement_test.go
```diff
@@ -534,3 +536,60 @@ func TestBindJSON(t *testing.T) {
...
+	createTable(t, db, `CREATE OR REPLACE TABLE arr_test(
+		arr_float float[2]
+	)`)
+
+	// Insert parameters.
+	_, err := db.ExecContext(ctx, `INSERT INTO arr_test VALUES (?)`,
+		[]float32{7, 1})
+	require.NoError(t, err)
+
+	prepared, err := db.Prepare(`SELECT list_value(?,?), list_cosine_distance(?,?)`)
```
taniabogatsch, line 554: Do `list_value` and `list_cosine_distance` trigger the `tryBind...` code, or do we know their parameter types?

statement_test.go
```diff
@@ -534,3 +536,60 @@ func TestBindJSON(t *testing.T) {
...
 }
+
+func TestPrepareComplexQueryParameter(t *testing.T) {
+	db := openDbWrapper(t, ``)
+	defer closeDbWrapper(t, db)
+
+	ctx := context.Background()
+	createTable(t, db, `CREATE OR REPLACE TABLE arr_test(
+		arr_float float[2]
+	)`)
```
taniabogatsch, line 547: Can we also add a test for nested lists/arrays? Something like `float[2][2]` or `varchar[][][]`.

crisismaple:
> Hi @crisismaple! Thanks a lot for your contribution! :) I just had a detailed look and left comments.

Hi @taniabogatsch, I just pushed a new commit, please help to review those changes ;)

statement.go
```diff
@@ -242,6 +240,13 @@ func (s *Stmt) bindJSON(val driver.NamedValue, n int) (mapping.State, error) {
 	return mapping.StateError, addIndexToError(unsupportedTypeError("JSON interface, need []byte or string"), n+1)
 }
 
+func (s *Stmt) bindUUID(val driver.NamedValue, t Type, n int) (mapping.State, error) {
```
taniabogatsch, line 243: Do we already have a test for binding UUIDs? If not, could you add one for this code path? 

And a small nit: we do not need to pass `t Type` here.

statement_test.go
taniabogatsch: Interesting! Thanks for clarifying :)

statement_test.go
```diff
@@ -534,3 +540,98 @@ func TestBindJSON(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, string(jsonData), str)
 }
+
+func TestPrepareComplexQueryParameter(t *testing.T) {
```
taniabogatsch, line 544: Can we add one more test that takes an empty slice as its input parameter?

taniabogatsch (changes requested):
Thanks for adding the changes! I've a few more minor comments, and then this PR should be ready to go!

taniabogatsch:
Great - could you just have a quick look at the tests, there's a build failure somewhere :)

crisismaple:
Just rebased main branch and fix build errors, all unit test should be fine now @taniabogatsch

PR #457, opened 2025-05-26
GetConnectionId and expose the BindData struct in the scalar UDF executor

taniabogatsch:
Expose `GetConnectionId`:
```Go
// GetConnectionId returns the connection ID of the internal DuckDB connection.
// It expects a *sql.Conn connection.
func GetConnectionId(c *sql.Conn) (uint64, error)
```

Adds a new struct `BindData`, which contains data DuckDB, and in the future also clients, can set during scalar function binding. Currently, this only contains the connection ID. Together with `GetConnectionId`, this enables users to map their scalar functions to the connection that executes the scalar function.

I'm opening this as a draft PR, as there are still a few unresolved points. Also, to make arbitrary binding data available during execution, I need to PR more C API functions to DuckDB main.

**Questions:**
- Currently, go-duckdb does not keep a mapping of its open connections. Also, they do not yet contain a state as proposed in #418. Should clients keep such a mapping, and use only the connection ID to retrieve the matching connection? Should we make the context available, in `Conn` or even `BindData`?
- I've extended `type ScalarFuncExecutor struct` with another field allowing a new type of executor. This should be in line with Go's compatibility rules (https://go.dev/blog/module-compatibility) for structs. But please let me know if I am missing something here.

cc @VGSML @lorenzowritescode

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -198,3 +198,20 @@ func (conn *Conn) prepareStmts(ctx context.Context, query string) (*Stmt, error)
 
 	return conn.prepareExtractedStmt(*stmts, count-1)
 }
+
+// GetConnectionId returns the connection ID of the internal DuckDB connection.
+// It expects a *sql.Conn connection.
+func GetConnectionId(c *sql.Conn) (uint64, error) {
+	var connId uint64
+
+	err := c.Raw(func(driverConn any) error {
+		conn := driverConn.(*Conn)
+		var ctx mapping.ClientContext
+		mapping.ConnectionGetClientContext(conn.conn, &ctx)
+		defer mapping.DestroyClientContext(&ctx)
+		connId = uint64(mapping.ClientContextGetConnectionId(ctx))
+		return nil
+	})
+
+	return connId, err
+}
--- a/scalar_udf.go
+++ b/scalar_udf.go
@@ -58,44 +58,33 @@ type scalarFuncInfo interface {
 	Config() ScalarFuncConfig
 }
 
-// ScalarFuncExecutor contains the row-based function to execute a user-defined scalar function.
+type (
+	// RowExecutorFn TODO.
+	RowExecutorFn func(values []driver.Value) (any, error)
+	// RowExecutorWithBindDataFn TODO.
+	RowExecutorWithBindDataFn func(bindData BindData, values []driver.Value) (any, error)
+)
+
+// ScalarFuncExecutor contains the functions to execute a user-defined scalar function.
+// It invokes the first non-nil member.
 type ScalarFuncExecutor struct {
 	// RowExecutor accepts a row-based execution function.
 	// []driver.Value contains the row values.
 	// It returns the row execution result, or error.
-	RowExecutor func(values []driver.Value) (any, error)
+	RowExecutor RowExecutorFn
+	// RowExecutorWithBindData accepts a row-based execution function.
+	// BindData contains any data set during the binding phase.
+	// []driver.Value contains the row values.
+	// It returns the row execution result, or error.
+	RowExecutorWithBindData RowExecutorWithBindDataFn
 }
 
-// ScalarFunc is the default row-based user-defined scalar function interface.
+// ScalarFunc is the user-defined scalar function interface.
 // Any scalar function must implement a Config function, and an Executor function.
 type ScalarFunc interface {
 	scalarFuncInfo
 	// Executor returns ScalarFuncExecutor to execute the scalar function.
 	Executor() ScalarFuncExecutor
-}
-
-// ScalarFuncExecutorWithBindData contains the row-based function to execute a user-defined scalar function.
-// It has access to any data set during the binding phase of the scalar function.
-type ScalarFuncExecutorWithBindData struct {
-	// RowExecutor accepts a row-based execution function.
-	// BindData contains any data set during the binding phase.
-	// []driver.Value contains the row values.
-	// It returns the row execution result, or error.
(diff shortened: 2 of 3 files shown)


VGSML:
Thanks a lot for the PR! It looks like it would be simpler to pass context.Context directly to the UDF instead of BindData. However, in that case, we would need to implement logic to store the Go context at the duckdb.Conn (duckdb.Connector) level.
I can propose a solution — could you create a separate branch here where I can propose my changes the top of yours (create PR)?

Also, we need to implement the same logic to retrieve the connection ID from the table UDF bind info.

taniabogatsch:
EDIT: I was writing this before you submitted your PR to my fork - will check that out now (thanks for the input!) :)

> It looks like it would be simpler to pass context.Context directly to the UDF instead of BindData.

What I like about having something like `BindData` is that its a more explicit way to allow users to pass arbitrary bind data to the execution (in the future). Although I guess that's also possible if we make the context available both during binding and execution. 🤔 I agree that the current way `BindData` looks in this PR should not be the final solution. Maybe it could contain a key-value map, which could then also contain the context?

Here's a bit more information on how I envision an (optional) bind callback in the future, once we have this PR available: https://github.com/duckdb/duckdb/pull/17666. For example, internally, during bind, we can have the input parameters available as `Expression`.
```cpp
typedef unique_ptr<FunctionData> (*bind_scalar_function_t)(ClientContext &context, ScalarFunction &bound_function,
                                                           vector<unique_ptr<Expression>> &arguments);
```
In the future, we can expose these in the C API, and make it possible to, e.g., have access to a constant expression value, or a return type. With that knowledge, the client can then store additional information (necessary for execution) in the `BindData`.

> Also, we need to implement the same logic to retrieve the connection ID from the table UDF bind info.

That's a good point, I haven't checked if we already have the necessary C API functions in place for that.

context.go
```diff
@@ -0,0 +1,64 @@
...
+	Context(connId uint64) context.Context
+	SetContext(connId uint64, ctx context.Context)
+	Remove(connId uint64)
+}
+
+// ctxStore is a thread-safe implementation of contextStore that uses a map to store contexts by connection ID.
+type ctxStore struct {
+	mu    sync.Mutex
+	store map[uint64]context.Context
+}
```
lorenzowritescode, line 25: tip: you can use a sync.Map instead of managing a lock

From https://pkg.go.dev/sync#Map

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

This will give you very large improvement on read performance - I have ran some experiments in the past and when you have a lot of readers, using this over a mutex can give you a 50x speedup

scalar_udf.go
```diff
@@ -42,12 +46,32 @@ type ScalarFuncConfig struct {
 	SpecialNullHandling bool
 }
 
-// ScalarFuncExecutor contains the callback function to execute a user-defined scalar function.
-// Currently, its only field is a row-based executor.
+// BindInfo TODO: How should this look like?
+// TODO: In future PRs, clients should be able to set their custom data here during `func Binder(BindInfo) any`,
+// TODO: so they can access is during execution.
+type BindInfo struct {
+	ConnId uint64
```
lorenzowritescode, line 53: If we are surfacing context directly, I doubt we need to surface `connId` too here - what do you think?

connection.go
```diff
@@ -59,6 +72,7 @@ func (conn *Conn) QueryContext(ctx context.Context, query string, args []driver.
 		return nil, err
 	}
 
+	conn.setContext(ctx)
```
lorenzowritescode, line 75: I think it would be wise to unset the context after this (and other) functions return. Ideally, context should be nil when the connection is idle

context.go
```diff
@@ -0,0 +1,64 @@
...
+
+// ctxConnIdKey is a key used to store the connection ID in the context.
+var ctxConnIdKey ctxConnIdKeyType = "duckdb_conn_id"
+
+// contextStore is an interface for managing connection contexts.
+type contextStore interface {
+	Context(connId uint64) context.Context
+	SetContext(connId uint64, ctx context.Context)
+	Remove(connId uint64)
+}
```
lorenzowritescode, line 19: I think this interface is unnecessary unless I am missing something. You can replace usages of this type with the concrete type *ctxStore.

This is just a go style thing, feel free to ignore. Comes from the accept interfaces, return structs concept explained here: https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b

lorenzowritescode (approved):
Thanks @taniabogatsch!

context.go
VGSML: I think in this case it’s better to use a regular mutex directly, because the use case involves a similar number of read and write operations. RWMutex and sync.Map are more efficient when read operations significantly outnumber writes. Otherwise, they can increase GC overhead.

connection.go
VGSML: I thought about, but if we will create a lot of new connections, that will change their Id's the map will grow and memory can leak. It's not a big numbers but in any case)

taniabogatsch: Done.

VGSML:
all good, you can merge it

PR #476, opened 2025-06-20
Arrow stream reader

VGSML:
Replace the arrow.simpleRecords arrow.RecordReader implementation to our own streaming implementation to avoid collecting all records in the memory.

closed #445


VGSML:
@taniabogatsch do you have time to review?

arrow.go
```diff
@@ -125,45 +217,28 @@ func (a *Arrow) QueryContext(ctx context.Context, query string, args ...any) (ar
 	if err != nil {
 		return nil, err
 	}
-	defer stmt.Close()
```
taniabogatsch, line 128: Do we need to move this down into the `err != nil` calls? If we `defer` here, we'll clean up there, no?

taniabogatsch (commented):
Hi @VGSML, finally got around to work on go-duckdb a bit again. I only have one comment/question, otherwise, the PR looks good to go to me. :)

taniabogatsch:
> Actually I propose to exclude this tests, because it has no sense to use the reader concurrently, instead of send results to some channel.

Maybe we can add a comment to the test instead? I'd say keeping it to test the functionality (i.e., more a unit test instead of a usage example) could still make sense. 🤔

arrow.go
VGSML: yes, your are right!

PR #483, opened 2025-07-05
Simplify code for table functions

JAicewizard:
This reduces the amount of code needed for table functions. This is not fully the case, as the linecount did go up, but at least 50 lines of that have to be in my appender PR anyways. As you can see the giant ugly switch statement is no longer present.

All code is now made purely for parallel functions only, with some code to wrap the non-parallel functions to make them "parallel" (with max 1 concurrent invocation).

### DIFF
--- a/table_source.go
+++ b/table_source.go
@@ -57,7 +57,7 @@ type (
 	ChunkTableSource interface {
 		sequentialTableSource
 		// FillChunk takes a Chunk and fills it with values.
-		// Set the chunk size to 0 for end the function.
+		// Set the chunk size to 0 to end the function.
 		FillChunk(DataChunk) error
 	}
 
@@ -70,7 +70,7 @@ type (
 	ParallelChunkTableSource interface {
 		parallelTableSource
 		// FillChunk takes a Chunk and fills it with values.
-		// Set the chunk size to 0 for end the function
+		// Set the chunk size to 0 to end the function
 		FillChunk(any, DataChunk) error
 	}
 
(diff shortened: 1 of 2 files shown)


JAicewizard:
@taniabogatsch I also have a question about the current bindings situation, maybe this is the wrong place to ask, but why is all the code split into different packages for each platform? I would not expect that each platform required different bindings (just linking with different libraries of course). (The intermediary module of go-duckdb/bindings might also have a performance impact, but I am not sure. Would have to look into it to be fully sure.)

PS. I would loved to have a bit more time next semester by sneaking in some time at a ddbl internship. Sadly this seems to have fallen through. I will redo the appender API, and see what else I can work on over the summer. go-duckdb seems to be in a pretty good state however, I don't think there would be much I could work on :D

taniabogatsch:
> All code is now made purely for parallel functions only, with some code to wrap the non-parallel functions to make them "parallel" (with max 1 concurrent invocation).

Sounds great, and the code looks much cleaner now!

> why is all the code split into different packages for each platform? I would not expect that each platform required different bindings (just linking with different libraries of course).

That's because of Go build constraints - I'd love for Go to have a way to import different modules based on OS + architecture, but that's not possible. So, instead, we only build certain files, and that file imports the package of the platform. We need (afaik) different packages because they directly (the pre-built ones) contain the `.a` files, which differ for each OS + architecture, so they can't be in one package (maybe I am wrong, but vendoring makes this at least annoying)? Or at least we would clutter that package with a ton of `.a` libraries we don't want there, as they are for different OS + architecture combinations.

PR #484, opened 2025-07-07
Improve appender speed v2

JAicewizard:
This currently also includes the changes of #483, but I will rebase once that is merged.

This adds multiple ways to append, using all the variations of table sources.
Using just row-based is actually slower than the current solution, which may be expected as it operates very similarly but with extra steps. The real performance gains come from the parallel implementations.
The benchmarks don't show much improvement (~25% for both parallel row and parallel chunk), however this is due to the fact that they are both bottlenecked on appending the chunks in DuckDB itself. If instead the bottleneck would be the data ingestion/computation on the user-side (instead of a simple counter), this would be much more favorable to the parallel variations. Currently they barely show up as parallel on my laptop.

Another improvement would be setting entire vectors of data at a time, instead of individual values, this could be done in a future PR.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -3,25 +3,42 @@ package duckdb
 import (
 	"database/sql/driver"
 	"errors"
+	"runtime"
+	"sync"
 
 	"github.com/marcboeker/go-duckdb/mapping"
 )
 
-// Appender holds the DuckDB appender. It allows efficient bulk loading into a DuckDB database.
-type Appender struct {
-	conn     *Conn
-	schema   string
-	table    string
-	appender mapping.Appender
-	closed   bool
+type (
+	appenderSource[T any] struct {
+		Source T
+	}
 
-	// The chunk to append to.
-	chunk DataChunk
-	// The column types of the table to append to.
-	types []mapping.LogicalType
-	// The number of appended rows.
-	rowCount int
-}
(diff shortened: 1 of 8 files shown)


JAicewizard:
Benchmarks for the new function:

```
BenchmarkAppenderNested-16                 	    100	 12083417 ns/op
BenchmarkAppenderSingle/int8-16            	    288	  4121898 ns/op
BenchmarkAppenderRowSingle/int8-16         	    217	  5464273 ns/op
BenchmarkAppenderParallelRowSingle/int8-16 	    391	  3000902 ns/op
BenchmarkAppenderChunkSingle/int8-16       	    378	  3183649 ns/op
BenchmarkAppenderParallelChunkSingle/int8-16         	    379	  3030166 ns/op
```

You may also notice that this is actually integrated into the already existing appender API, instead of what I did before. The current appender implementation was so similar to what I was making, it would be a waste of code to duplicate all of that.

mlafeldt:
👋 Thanks for your contribution!

This repository has moved to https://github.com/duckdb/duckdb-go

If this PR is still relevant, please re-open it at the new location. I'm closing this now as part of the migration.

Please see https://github.com/marcboeker/go-duckdb/issues/565 for more details.

PR #486, opened 2025-07-08
fix: handle nil pointers in nullable UUID column (fixes #485)

### LINKED ISSUE #485: Inserting NULL UUID fails in v2.3.1
Prior to v2.3.1, inserting a NULL value nil UUID ptr, either with internal type *UUID or google *uuid.UUID worked well.

However, starting in v2.3.1, inserting a nil UUID ptr fails with:
```
panic: value method github.com/google/uuid.UUID.String called using nil *UUID pointer
```

This happens because `bindUUID()` fails to check if the value is nil before calling `ss.String()`.

kfiles-snyk:
Handles NULL values in UUID columns. 
Adds test coverage for a variety of UUID cases.

Issue: https://github.com/marcboeker/go-duckdb/issues/485

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"reflect"
 
 	"github.com/marcboeker/go-duckdb/mapping"
 )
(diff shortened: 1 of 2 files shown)


taniabogatsch:
Hi @kfiles-snyk, thanks for the PR - I only saw a few nits and went ahead to quickly push them. Hope you don't mind. :) Will merge this once the tests have passed.

Issue #480, opened 2025-06-30
Have the query parameter binding rules of go-duckdb been changed?


tnamao:
Hello,

When binding parameters in queries using go-duckdb, I encountered a discrepancy in behavior compared to previous versions.
When binding an int64 pointer to a query, versions 2.3.1 and 2.3.2 fail because they cannot find a corresponding type, resulting in an error.
Has this behavior been changed intentionally?

### Reproduction Code

```go
package main

import (
	"database/sql"
	"fmt"

	_ "github.com/marcboeker/go-duckdb/v2"
)

func main() {
	db, err := sql.Open("duckdb", "")
	if err != nil {
		panic(err)
	}
	defer db.Close()

	// Create a table
	_, err = db.Exec("CREATE TABLE IF NOT EXISTS users (id BIGINT, name VARCHAR)")
	if err != nil {
		panic(err)
	}

	var idRef *int64
	id := int64(1862483)
	idRef = &id
	_, err = db.Exec("INSERT INTO users (id, name) VALUES (?, ?)", idRef, "Alice")
	if err != nil {
		fmt.Printf("Error inserting user: %v\n", err)
		panic(err)
	}

	fmt.Println("successfully inserted user")
}
```

### Results

- go-duckdb 2.3.1, 2.3.2 (error)

```
Error inserting user: could not bind parameter
unsupported data type: unknown type: index: 1
```

- go-duckdb 2.3.0 (success)

```
successfully inserted user
```

tnamao:
https://github.com/marcboeker/go-duckdb/pull/493
I sent a pull request to add test code.
Please review it.

I have confirmed that the fix has been applied on my end, so please close the issue when convenient.

Thank you!

PR #487, opened 2025-07-08
fix: correctly insert pointer values

kfiles-snyk:
Fixes the handling of values of pointer type. These should be passed on to the DefaultParamterConverter, as go-duckdb doesn't natively handle most such types. This includes pointers to types which implement driver.Valuer and fmt.Stringer.

Also adds a Valuer interface implementation to UUID type to enable UUID pointers to be converted.

Issue: https://github.com/marcboeker/go-duckdb/issues/480


types.go
```diff
@@ -59,6 +61,11 @@ func (u *UUID) String() string {
 	return string(buf)
 }
 
+// Value implements the driver.Valuer interface.
+func (u *UUID) Value() (driver.Value, error) {
+	return u.String(), nil
+}
+
```
taniabogatsch, line 68: Can we also add a test for this? Or I might be missing something and we already have a test hitting it? Your added test does not hit the breakpoint for me 🤔

statement_test.go
```diff
@@ -651,3 +651,34 @@ func TestPrepareComplexQueryParameter(t *testing.T) {
...
+func TestInsertCustomTypePtr(t *testing.T) {
+	db := openDbWrapper(t, ``)
+	defer closeDbWrapper(t, db)
+
+	// Create a table with a VARCHAR column
+	createTable(t, db, `CREATE TABLE test_custom_type_ptr(id INTEGER, name VARCHAR)`)
+
+	// Create a value of our custom type
+	expected := "test string"
+	customValue := String256(expected)
```
taniabogatsch, line 668: Don't we expect the custom type, here, too, instead of falling back to `string`?
I.e., 
```Go
expected := String256("test string")
```
and
```Go
var actual String256
```

types.go
kfiles-snyk: It's tested by the TestBindUUID patch. Now that you've merged that PR, I've rebased and the `Value()` function is covered by the test.

taniabogatsch:
Thanks!

PR #495, opened 2025-07-14
Use proper SetRowValue to handle unprojected columns

wmTJc9IK0Q:
row.SetRowValue was not properly changing the column index when a previous column is not projected. I am not sure why we have both of these methods but I am keeping them both to avoid breaking changes.


taniabogatsch:
Thanks!

PR #503, opened 2025-07-28
FIX: Memory leak in arrow interface{}

VGSML:
I have forgot to make release for last arrow.Record in the next() reader func.


VGSML:
Thanks, @taniabogatsch Can you make release version with this and arrow v18.4.0?

PR #510, opened 2025-08-10
fix: invalidation of cache on connection close

### LINKED ISSUE #509: InstanceCache is not invalidated during Connector.close(), preventing reconnection to the same database.
Hello,
I observed an issue where trying to reopen a connection to a previously accessed database causes the `sql.Open()` function to hang. The situation looks like the following:
1. Connection was opened to `x.db`
2. Connection to `x.db` was closed.
3. Connection was opened to `y.db`.
4. Connection to `y.db` was closed.
5. Connection to `x.db` was supposed to open considering that I've closed it before but it doesn't. It just hangs indefinitely.

On delving deeper into the issue, I noticed that this originates from the InstanceCache not being invalidated. I suppose that the stale cache assumes that the connection already exists but cannot find it hence causes it to malfunction. On reading through the [go bindings lib](https://github.com/duckdb/duckdb-go-bindings/blob/main/bindings.go#L800), the comments above `CreateInstanceCache` state that the cache should be destroyed with `DestroyInstanceCache`.

This function was missing from the implementation for the library, which I think is the cause of my issue. I'm not entirely sure about my solution but I propose adding in an invalidation function too which could be called when the connection is closed?

Here are the references to the code blocks where:
- [The connection is opened with the cache being created](https://github.com/marcboeker/go-duckdb/blob/71d80ce34da7a6339ebb620ca8139edccbc781d3/duckdb.go#L91)
- [The singleton cache which doesn't update](https://github.com/marcboeker/go-duckdb/blob/71d80ce34da7a6339ebb620ca8139edccbc781d3/duckdb.go#L19)
- [Where closing the connection should invalidate the cache too](https://github.com/marcboeker/go-duckdb/blob/71d80ce34da7a6339ebb620ca8139edccbc781d3/duckdb.go#L132)

I'm new to this so perhaps my solution may not be the best, but I'll raise a PR open to critique.
Thanks for creating and maintaining the library!

dat-adi:
Due to the absence of cache invalidation, we cannot re-establish a connection to a previously connected database. This commit serves to fix this by adding in the required functionality.

This resolves #509

### DIFF
--- a/duckdb.go
+++ b/duckdb.go
@@ -22,7 +22,7 @@ var (
     cacheCreated  bool
 )
 
-func GetInstanceCache() mapping.InstanceCache {
+func getInstanceCache() mapping.InstanceCache {
 	cacheMutex.Lock()
 	defer cacheMutex.Unlock()
 	
@@ -33,7 +33,7 @@ func GetInstanceCache() mapping.InstanceCache {
 	return instanceCache
 }
 
-func InvalidateInstanceCache() {
+func destroyInstanceCache() {
 	cacheMutex.Lock()
 	defer cacheMutex.Unlock()
 	
@@ -110,7 +110,7 @@ func NewConnector(dsn string, connInitFn func(execer driver.ExecerContext) error
 		state = mapping.OpenExt("", &db, config, &errMsg)
 	} else {
 		// Open a file-backed database.
-		state = mapping.GetOrCreateFromCache(GetInstanceCache(), getDBPath(dsn), &db, config, &errMsg)
+		state = mapping.GetOrCreateFromCache(getInstanceCache(), getDBPath(dsn), &db, config, &errMsg)
 	}
 	if state == mapping.StateError {
 		mapping.Close(&db)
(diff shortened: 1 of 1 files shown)


taniabogatsch (commented):
Hi @dat-adi, thanks for spotting this and proposing a fix! I've left two small nits, but the approach looks good to me :) I'll approve running the tests on it, too!

dat-adi:
I considered adding in a atomic reference counter and destroy based on the last connection object, but I think that would only recreate the initial problem of the lack of invalidation of stale cache. Do you guys have documentation for setting up testing on a local system?

I want to run by my approaches on local since it might be expensive otherwise.

mlafeldt:
👋 Thanks for your contribution!

This repository has moved to https://github.com/duckdb/duckdb-go

If this PR is still relevant, please re-open it at the new location. I'm closing this now as part of the migration.

Please see https://github.com/marcboeker/go-duckdb/issues/565 for more details.

PR #525, opened 2025-09-09
fix: add RecordBatch to support arrow v18.4.1

### LINKED ISSUE #524: incompatible with arrow v18.4.1
Arrow v18.4.1 introduced a breaking change in the `array.RecordReader` interface, as they are deprecating `Record` for `RecordBatch`

`/github.com/marcboeker/go-duckdb/v2@v2.3.6/arrow.go:237:9: cannot use &arrowStreamReader{…} (value of type *arrowStreamReader) as array.RecordReader value in return statement: *arrowStreamReader does not implement array.RecordReader (missing method RecordBatch)`

derekperkins:
the upstream arrow library introduced a breaking change in a patch version, by changing the interface functions. They are deprecating Record in favor of RecordBatch, which is a naming only change at this point. There are type aliases added, but that doesn't help fulfilling the new interface

fixes #524


zeroshade (approved):
Sorry about the complication here!

PR #530, opened 2025-09-10
Fix instance cache key collisions with different configurations

mlafeldt:
## Summary

This PR fixes instance cache key collisions that occur when connecting to the same database path with different authentication or session configurations. 

**Problem:** The current cache key generation only uses the database path (via `getDBPath(dsn)`), ignoring query parameters. This causes cache collisions when multiple users connect to MotherDuck with different `motherduck_token` and `motherduck_session_hint` values, leading to incorrect connection sharing.

**Solution:** Enhanced the cache key generation to include relevant configuration parameters:
- Added configurable list of cache-relevant parameters (`motherduck_token`, `motherduck_session_hint`, `auth_token`, `user`, `session_id`)
- Hash configuration values to create unique cache keys per configuration combination  
- Maintains backward compatibility for connections without cache-relevant parameters

## Changes

- **New `getCacheKey()` function**: Replaces `getDBPath(dsn)` in cache key generation (duckdb.go:104)
- **Configurable parameters**: `cacheRelevantParams` slice defines which parameters affect caching
- **Consistent hashing**: Sorted parameter pairs ensure consistent cache keys
- **Comprehensive tests**: Added `cache_key_test.go` with uniqueness, consistency, and edge case tests

## Test Plan

✅ All existing tests pass  
✅ New cache key tests verify:
- Correct handling of DSNs without cache-relevant parameters
- Unique keys for different configurations
- Consistent keys for identical configurations  
- Proper handling of mixed cache-relevant and irrelevant parameters

## Impact

- **MotherDuck**: Enables proper multi-user isolation with different tokens/sessions
- **Airport extension**: Supports different `auth_token` values  
- **Other extensions**: Framework for cache key customization
- **Backward compatibility**: No changes for existing single-user scenarios

## Addresses

- https://github.com/duckdblabs/motherduck/issues/359
- Related to resolved DuckDB core issue via https://github.com/duckdb/duckdb/pull/13129

## Testing

```bash
go test -v . -run TestGetCacheKey
go test -v . -run TestCacheKey
go test -v .
```

All tests pass with the implementation.


duckdb.go
```diff
@@ -144,6 +145,48 @@ func getDBPath(dsn string) string {
 	return dsn[0:idx]
 }
 
+// Configuration parameters that should affect cache keys. These are authentication
+// and session parameters that distinguish different connections to the same database.
+var cacheKeyParams = []string{
```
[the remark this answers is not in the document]
taniabogatsch: From my understanding of the issue, users try to pass a unique token here, which is then used to disambiguate two connections (whose parameters otherwise overlap). So maybe it is better to make this completely configurable by the user, without providing any pre-configured values?

duckdb.go
```diff
@@ -144,6 +145,48 @@ func getDBPath(dsn string) string {
...
+
+	if len(relevantParams) == 0 {
+		return basePath
+	}
+
+	// Creates a sorted list of key-value pairs for consistent hashing
+	paramString := relevantParams.Encode()
+	hash := sha256.Sum256([]byte(paramString))
+
+	return fmt.Sprintf("%s#%x", basePath, hash[:8]) // Use first 8 bytes of hash
```
taniabogatsch, line 187: Just for my understanding: why only the first 8 bytes of the hash?

taniabogatsch, line 187: Also, how does the `#` work here? If we use, e.g., `hello.db#myhash` as the DB path, isn't that going to create unique new files with the different hashes instead of, e.g., opening an existing file? I tried to have a look at the C++ side of this: `DBInstanceCache::GetOrCreateInstance` calls into either `CreateInstanceInternal` or `GetInstanceInternal`, both of which call `GetDBAbsolutePath(database, *local_fs);` on the `database` string (path). I can't find any special-handling of the `#`, but maybe I missed it / miss something here?

PR #536, opened 2025-09-13
perf(rows): precompute fixed reflect types

ayuhito:
Instead of calling `reflect.TypeOf(type{})` repetitively, we can precompute these types to save some CPU cycles and potentially skip an extra allocation for any non-primitive types (e.g. time.Time).


taniabogatsch (approved):

taniabogatsch:
Thanks!

PR #546, opened 2025-09-18
perf(rows): use strings builder types

ayuhito:
Not very impactful, but why not. Each concat is a separate heap allocation that is discarded by GC, so `strings.Builder` is a little bit more hygienic for this use case especially since we are appending in an unknown loop.


taniabogatsch:
Thanks! Reducing heap allocations is great - even though, like you said, this is probably not the most performance-critical code path.

PR #550, opened 2025-09-22
Expose scalar UDF arguments

taniabogatsch:
DuckDB `v1.4` exposes parts of the scalar function binding interface, including a lightweight API around expressions.

With that, we can make it possible for clients implementing their custom scalar UDFs to also provide a custom bind callback. Binding happens *before* scalar function execution, meaning any computation based on the input arguments _that only has to happen once_, should happen during that bind step. Then, the result of that bind step can be made available to the different threads running the "actual" scalar function callback during execution.

In this PR, I've drafted an initial interface to make that possible. There are still a few open questions, like:
- should the custom bind data (currently a returned `ScalarUDFArg`) be set in the context, or should we add (yet) another row executor taking that argument as its input? Or, should we add another row executor, but hand it some info-struct, to allow more expansion in the future. I think we want to discuss the options here, as this is also relevant for a long-term stable interface.
- does this PR sufficiently meet requirements on custom bind callbacks? Is there anything we "forget" by not exposing expressions and their API directly, and instead wrapping them in `ScalarUDFArg`?

### DIFF
--- a/scalar_udf.go
+++ b/scalar_udf.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"database/sql"
 	"database/sql/driver"
+	"fmt"
 	"runtime"
 	"runtime/cgo"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


lorenzowritescode (approved):
Thanks Tania, looks great - this will help us simplify our usage of UDFs!

taniabogatsch:
Will close this here and re-open it in duckdb/duckdb-go!

PR #563, opened 2025-10-11
Optimize with reflect.TypeFor

apocelipes:
To simplify code and to avoid repeated allocations.

### DIFF
--- a/types.go
+++ b/types.go
@@ -21,28 +21,30 @@ import (
 
 // Pre-computed reflect type values to avoid repeated allocations.
 var (
-	reflectTypeBool      = reflect.TypeOf(true)
-	reflectTypeInt8      = reflect.TypeOf(int8(0))
-	reflectTypeInt16     = reflect.TypeOf(int16(0))
-	reflectTypeInt32     = reflect.TypeOf(int32(0))
-	reflectTypeInt64     = reflect.TypeOf(int64(0))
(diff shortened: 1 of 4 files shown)


types.go
```diff
@@ -21,28 +21,30 @@ import (
...
+	reflectTypeString         = reflect.TypeFor[string]()
+	reflectTypeBytes          = reflect.TypeFor[[]byte]()
+	reflectTypeDecimal        = reflect.TypeFor[Decimal]()
+	reflectTypeSliceAny       = reflect.TypeFor[[]any]()
+	reflectTypeMapString      = reflect.TypeFor[map[string]any]()
+	reflectTypeMap            = reflect.TypeFor[Map]()
+	reflectTypeUnion          = reflect.TypeFor[Union]()
+	reflectTypeAny            = reflect.TypeFor[any]()
+	reflectTypeUUID           = reflect.TypeFor[UUID]()
+	reflectTypeMappingHugeInt = reflect.TypeFor[mapping.HugeInt]()
```
taniabogatsch: nit: let's call this `reflectTypeHugeInt`

taniabogatsch (changes requested):
Thanks! Minor nit and then this looks good to go!

types.go
apocelipes: Thanks for review! Done.

taniabogatsch (approved):

taniabogatsch:
Thanks!

PR #564, opened 2025-10-15
Announce project move to duckdb/duckdb-go

mlafeldt:
Depends on https://github.com/duckdb/duckdb-go/pull/4

We need to push a new tag to make the deprecation notice take effect.

> To deprecate a module, an author may add a // Deprecated: comment and tag a new release.

https://go.dev/ref/mod


taniabogatsch:
Since we plan to migrate issues and discussions gracefully with more thorough triaging, we probably want to keep the repository out of readonly for a few more days after merging this and pushing the tag (cc @marcboeker). :)

taniabogatsch (approved):

